PR #45, opened 2026-02-08
i18n: add French translation

UYTR5:
Adds French (fr) translations to Localizable.xcstrings.


stonerl (approved):
Perfect, thank you. I added a new string recently that you might not have translated. But you can do so later.

stonerl:
Just noticed. You translated it. Nice work.

PR #52, opened 2026-02-09
german is now 100% translated

Mutzki13:
The missing entries from the last update are now translated into german, too.


BillChirico:
@Mutzki13, could you please add more information to the description?

stonerl (approved):
Nice. Tausend Dank. 🙏

PR #54, opened 2026-02-10
fix: prevent CheckedContinuation double-resume crash in MenuBarItemSpacingManager

7a6163:
## Summary

  - Fix a race condition in `signalAppToQuit` where both the KVO observer and the failsafe timeout could call `continuation.resume()`, causing a runtime crash
  (`CheckedContinuation` traps on double-resume)
  - Add `OSAllocatedUnfairLock<Bool>` guard to ensure only the first path to arrive actually resumes the continuation
  - Remove redundant `Task.isCancelled` check, as the lock now handles all race scenarios reliably

  ## Details

  In `MenuBarItemSpacingManager.signalAppToQuit`, a `withCheckedThrowingContinuation` has two resume paths:

  1. **KVO path** — fires when `app.isTerminated` becomes `true`
  2. **Failsafe path** — fires 1 second after `forceTerminate()` if KVO never fires

  These two paths could race, resulting in a double-resume crash. The fix introduces an atomic flag (`OSAllocatedUnfairLock<Bool>`) so that whichever path arrives first
  performs the resume, and the second is safely skipped.

  ## Test plan

  - [ ] Launch Thaw, trigger a spacing reset that requires quitting and relaunching a menu bar app
  - [ ] Verify the app quits and relaunches without crash
  - [ ] Force a slow-to-terminate app scenario (e.g., unresponsive process) to exercise the failsafe path

### DIFF
--- a/Thaw/MenuBar/Spacing/MenuBarItemSpacingManager.swift
+++ b/Thaw/MenuBar/Spacing/MenuBarItemSpacingManager.swift
@@ -8,6 +8,7 @@
 
 import Cocoa
 import Combine
+import os.lock
 import OSLog
 
 /// Manager for menu bar item spacing.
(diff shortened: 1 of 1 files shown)


BillChirico:
Great find and fix!

stonerl (approved):

PR #58, opened 2026-02-10
i18n: add zh-Hans translation

picpo:


stonerl (approved):
Nice work, thank you very much.

PR #76, opened 2026-02-11
feat: add Thai translation

priesdelly:
Add Thai translation.


priesdelly:
I need resolve the conflict first..

PR #89, opened 2026-02-14
perf: cache menu bar item windows to avoid Window Server queries in draw()

7a6163:
## Summary
  - Cache `getMenuBarItemWindows()` results instead of querying the Window Server synchronously on every `draw(_:)` call in `MenuBarOverlayPanelContentView`
  - Cached item windows are updated event-driven via existing publishers (`controlItem.$onScreenFrame` and `$applicationMenuFrame`) and populated once on initialization
  - Eliminates expensive, blocking IPC calls from the rendering hot path, which previously could fire dozens of times per second during active menu bar changes

  ## Details

  `pathForSplitShape()` (called from `draw(_:)`) was issuing a synchronous `Bridging.getMenuBarWindowList` IPC call to the Window Server on every redraw. Multiple
  high-frequency publishers trigger `needsDisplay = true`, causing this query to run repeatedly on the main thread.

  ### Changes
  - Added `cachedItemWindows: [WindowInfo]` property
  - Added `updateCachedItemWindows()` method
  - Cache is updated when:
    - Control item frame changes (`$onScreenFrame`) — covers show/hide transitions
    - Application menu frame changes (`$applicationMenuFrame`) — covers items added/removed by other apps (~10s interval)
    - On initialization — ensures first render has data
  - `pathForSplitShape()` now reads from cache instead of issuing a live query

  ### Thread safety
  All reads and writes to `cachedItemWindows` occur on the main thread. The array is a value type, so the local copy in `draw(_:)` is safe from concurrent mutation.

  ## Test plan
  - [ ] Split shape menu bar appearance renders correctly on first launch
  - [ ] Shape updates correctly when showing/hiding menu bar sections
  - [ ] Shape adjusts when other apps add/remove status bar items
  - [ ] Works on multi-monitor setup


stonerl:
Nice find, thank you.

stonerl (approved):

PR #97, opened 2026-02-14
perf: debounce and cancel redundant image cache updates

7a6163:
## Summary

  - Debounce merged image cache update publishers by 200ms to batch rapid-fire events (space change, screen change, color change, item cache change) into a single update
  - Cancel any in-flight `updateCache()` task before starting a new one, preventing redundant `CGWindowListCreateImage` calls from running concurrently
  - Add `Task.isCancelled` checks inside `updateCacheWithoutChecks` so cancelled tasks exit early — before expensive per-section screen captures and before applying stale results to the
  cache

  ## Context

  The image cache `configureCancellables` merges four publishers that can fire in quick succession (e.g. switching spaces triggers both space change and item cache change). Without
  debounce, each event independently triggered a full screen capture cycle. Without cancellation, overlapping captures could run in parallel, wasting CPU on duplicate
  `CGWindowListCreateImage` system calls.

  ## Test plan

  - [ ] Open the menu bar popup panel/search, verify item images still load correctly
  - [ ] Switch spaces rapidly, confirm no visual glitches or stale images
  - [ ] Monitor CPU usage in Activity Monitor during rapid events — should see reduced spikes compared to `main`
  - [ ] Open Settings → Menu Bar Layout, verify images render correctly after drag-and-drop reorder


stonerl (approved):

stonerl:
Thank you.

PR #102, opened 2026-02-15
perf: skip expensive timer work when panels are not visible

7a6163:
## Summary

  - Skip `updateWindowImage` screen capture in `IceBarColorManager` when IceBar panel is not visible, avoiding `CGWindowListCreateImage` calls every 5 seconds while idle
  - Skip overlay panel timer-based polling (30s wallpaper, 10s app menu frame) when panel is not on the active space, avoiding unnecessary Window Server queries
  - Stop permission polling timer (1s interval) once permission is granted, eliminating 60 checks/minute that serve no purpose after startup

  ## Context

  Multiple timers run continuously regardless of whether their associated UI is visible:

  | Timer | Interval | Expensive operation | When needed |
  |-------|----------|-------------------|-------------|
  | IceBar color | 5s | Screen capture + color calculation | Only when IceBar is shown |
  | Overlay wallpaper | 30s | Desktop wallpaper capture | Only on active space |
  | Overlay app menu | 10s | Window Server frame query | Only on active space |
  | Permission check | 1s | Permission API call | Only until granted |

  Combined, these produce ~80+ timer events/minute even when idle. With these guards, the timers still fire but their callbacks return immediately when the work is not needed.

  Each guard is backed by a separate event-driven path that ensures no updates are missed on state transitions:
  - IceBar calls `updateAllProperties` before showing the panel
  - Space changes trigger `activeSpaceDidChangeNotification` → full overlay update
  - `waitForPermission()` re-creates the timer if needed

  ## Test plan

  - [ ] Open IceBar, verify colors update correctly; close IceBar, confirm no screen capture activity in Console.app (`MenuBarOverlayPanel` / `IceBarColorManager` logs)
  - [ ] Switch between Spaces, verify overlay panel appearance updates on arrival
  - [ ] Change desktop wallpaper, verify overlay picks it up within 30s (while on active space)
  - [ ] Grant Accessibility/Screen Recording permissions on fresh install, verify app proceeds normally and timer stops (check with `os_log` or breakpoint)
  - [ ] Monitor CPU usage in Activity Monitor — idle CPU should drop compared to `main`


stonerl (approved):
Thank you. This is indeed something I wanted to look into as well.

Issue #63, opened 2026-02-11
[Bug]: Artifact on the second screen desktop


mittal-gunjit:
### Search for Similar Reports

- [x] I have searched existing issues for similar reports

### Description

<img width="1856" height="1172" alt="Image" src="https://github.com/user-attachments/assets/6ba4cb75-7037-4287-b537-f00c7f6edc07" />
I am using a secondary monitor (27 inch dell monitor) along with my macbook. this happens on only the secondary monitor
Here we are seeing this sort of a bar floating in the middle of screen. This is an issue i had experienced with ice as well.

I am using split menu bar with curved on the inward sides and block on the outside ends. I am not using thaw bar but instead only using the scroll to hide feature to better manage my menu bar.

also noticed that switching off   Use inset shape on screens with notch  option resolves this issue. 

https://github.com/user-attachments/assets/6b4a6152-8ed5-44bc-9a35-13732abac98f

### Steps to Reproduce

Configuration - I am using split menu bar with curved on the inward sides and block on the outside ends. I am not using thaw bar but instead only using the scroll to hide feature to better manage my menu bar. Also visible in the recording shared.

1) Have thaw running with the configuration described (not sure if it occurs with other configurations)
2) Go to a fullscreen app on your macbook
3) Go to a fullscreen app on secondary monitor
4) Go back to desktop on secondary monitor (macbook should still be on fullscreen app)

I was able to reproduce it by doing this. happy to provide others details on my configuration if required

### App Version

1.0.1-beta.6

### macOS Version

26.2

### Monitor Setup

Multi-monitor Setup

### Additional Information

Secondary dell 27 inch monitor along with my macbook in extended mode

stonerl:
Thank you for reporting. I'll look into this.

stonerl:
Can you check if this still happens with beta.7?

mittal-gunjit:
Yes it is still present for beta.7

mittal-gunjit:
Hi @stonerl , Have come up with a small fix on my own (with a lot of help from antigravity). If you can take a look it would be appreciated.

stonerl:
Looks good to me. Thank you.

mforell:
Did you manage to fix this? I have the same error when using multiple monitors.  

<img width="1227" height="326" alt="Image" src="https://github.com/user-attachments/assets/53a2aa92-4731-4397-a71f-dd71d659beea" />

stonerl:
Please update to RC4 and check if this still happens.

[...]

PR #123, opened 2026-02-16
Second screen artifact bug fix

mittal-gunjit:
## What does this PR do?

I'd appreciate it if you could give me a brief description of the changes proposed in this pull request.

## PR Type

- [✔️ ] Bugfix

- [ ] Yes
- [✔️ ] No

What is happening now?
The option "Use inset shape on screens with notch" results in a rendering issue where the "inset" menu bar shape is incorrectly applied to secondary screens, causing visual artefacts. The root cause is that NSScreen.hasNotch returns true for any screen with a non-zero top safe area (which includes standard screens with a menu bar), causing the app to apply notch-specific insets to standard displays.
Issue Number: 63

## What is the new behavior?
Update hasNotch computed property to use auxiliaryTopLeftArea instead of safeAreaInsets, which has fixed the issue.
What does this PR change or add?

## PR Checklist

- [✔️ ] I’ve built and run the app locally
- [✔️ ] I’ve checked for console errors or crashes
- [ ✔️] I've run relevant tests and they pass

## Other information

#63 This is the issue being targeted here. The fix is a rather simple change, in my opinion, though I am not sure how it will affect the overall app. Have built and run it on my machine, and it seems to fix this issue.

### Notes for reviewers

I am very new to this (contributing to open source), and frankly speaking, I have never even read Swift code before. This fix comes totally from antigravity, and hence I am not sure what it will break, but I have done basic compile and run and confirmed it fixes my issue, though I will appreciate if someone else can also replicate my findings as it is also my first time with xcode and testing the app while also having the app installed threw me for a loop.


stonerl (approved):
LGTM

PR #146, opened 2026-02-18
fix: eliminate data race in image cache saveToDisk

7a6163:
## Summary

  - Snapshot `images` dictionary on the calling thread before entering the detached background task
  - Previously, `Task.detached` captured `[weak self]` and accessed `self.images` from a background thread while the main thread could simultaneously mutate it — a read/write
  data race that could cause crashes or corrupted reads
  - Also removes the unnecessary `[weak self]` capture since the task now operates entirely on the snapshot

  ## Files changed

  - `Thaw/MenuBar/MenuBarItems/MenuBarItemImageCache.swift` — `saveToDisk()`

  ## Test plan

  - [ ] Build and run the app
  - [ ] Trigger cache save (activate/deactivate app several times)
  - [ ] Verify disk cache is written correctly (check `~/Library/Caches/com.stonerl.Thaw/`)
  - [ ] Run with Thread Sanitizer enabled in Xcode to confirm no data race warnings

### DIFF
--- a/Thaw/MenuBar/MenuBarItems/MenuBarItemImageCache.swift
+++ b/Thaw/MenuBar/MenuBarItems/MenuBarItemImageCache.swift
@@ -114,10 +114,10 @@ final class MenuBarItemImageCache: ObservableObject {
 
         guard let url = Self.cacheFileURL else { return }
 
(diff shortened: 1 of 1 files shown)


stonerl (approved):
Good catch.

PR #152, opened 2026-02-19
feat(i18n): Add complete Brazilian Portuguese localization

josericardo-fo:


stonerl (approved):
Wow, thank you very much. 🙏

PR #171, opened 2026-02-21
fix(icebar): eliminate show delay and false "No items" message

7a6163:
## Summary

  - Remove unnecessary `async` from `IceBarPanel.show()` and the `Task {}`
    wrapper in `MenuBarSection.show()`, eliminating a scheduling delay that
    caused the panel to appear ~1 frame late when clicking empty menu bar space
  - Add three-stage loading UX across all error branches to prevent misleading
    messages:
    - 0–600ms: "Loading menu bar items…" with spinner
    - 600ms–2.6s: status message with spinner (still retrying)
    - After 2.6s: final message with "Check permissions" link to settings
  - Add `Task.isCancelled` checks between background cache steps to skip
    unnecessary work after the panel is closed
  - Extract shared `openPermissionsSettings()` helper to eliminate duplicated
    button actions across three error branches
  - Add `.onChange(of: loadingTimedOut)` diagnostic logging to all error
    branches for better state transition tracking

  ## Test plan

  - [ ] Click empty menu bar area — Thaw Bar should appear instantly (no delay)
  - [ ] Open Thaw Bar rapidly — should never flash "No items in this section"
  - [ ] Revoke screen recording permission → open Thaw Bar → after ~3s should
        show "Check permissions" link instead of infinite spinner
  - [ ] Open and immediately close Thaw Bar — background cache task should
        cancel cleanly (no stale work)
  - [ ] Verify rehide timer still works correctly after removing `Task {}` wrapper

  ## Localization

  New string added: `"Check permissions"` — zh-Hant translation included (`檢查權限`).
  Other languages will fall back to English until translations are added.


stonerl (approved):
Amazing work thank you.

PR #187, opened 2026-02-22
fix: remove IceBar log spam and skip control item capture

CamilleGuillory:
## Summary

- Remove `diagLog.debug` calls from `contentHeight`, `itemMaxHeight`, and `targetSize` in `IceBar.swift` — these SwiftUI computed properties fire ~80 times per render pass with identical values, flooding the diagnostic log
- Filter out Thaw's own control items before image capture in `MenuBarItemImageCache` — they always capture as transparent via `CGWindowListCreateImage`, causing a perpetual fail -> blacklist -> cooldown -> retry cycle
- Remove unused `liveHeight` variable that was only referenced in the removed debug logs

## Test plan

- [x] Open the Ice Bar by clicking empty space — verify items display correctly
- [x] Check diagnostic logs are significantly quieter (no `contentHeight`/`itemMaxHeight`/`targetSize` spam)
- [x] Verify no "captured image is transparent for ControlItem.Visible" messages in logs
- [x] Test on a notched Mac if available to confirm `contentHeight` inset path still works

### DIFF
--- a/Thaw/MenuBar/IceBar/IceBar.swift
+++ b/Thaw/MenuBar/IceBar/IceBar.swift
@@ -317,13 +317,9 @@ private struct IceBarContentView: View {
 
     private var contentHeight: CGFloat {
         let menuBarHeight = screen.getMenuBarHeightEstimate()
-        let liveHeight = screen.getMenuBarHeight()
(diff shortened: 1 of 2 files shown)


stonerl (approved):

stonerl:
Thank you for the cleanup.

PR #189, opened 2026-02-22
perf(icebar): skip input pause when clicking temporarily shown items

CamilleGuillory:
## What does this PR do?

Eliminates a redundant `waitForUserToPauseInput()` delay when clicking menu bar items from the Ice Bar, reducing perceived latency by 100-500ms+.

## PR Type

- [ ] Bugfix
- [ ] CI/CD related changes
- [ ] Documentation
- [ ] Feature
- [ ] Refactor
- [x] Performance improvement
- [ ] Test addition or update

- [ ] Yes
- [x] No

When clicking an item in the Ice Bar, `temporarilyShow()` moves the item to the visible section, waits for its position to settle (up to 250ms), adds a 25ms processing buffer, then calls `click()`. Inside `click()`, `waitForUserToPauseInput()` polls every 50ms until the mouse has been inactive for 50ms — adding another 100-500ms+ of delay that is redundant since the position has already stabilized.

## What is the new behavior?

`click()` now accepts a `skipInputPause` parameter (defaulting to `false` so all existing callers are unaffected). `temporarilyShow()` passes `skipInputPause: true` when clicking, bypassing the redundant input pause. This mirrors the existing pattern used by `move()`, which already accepts and uses `skipInputPause`.

## PR Checklist

### Notes for reviewers

The change is minimal — two call sites in `MenuBarItemManager.swift`. The `skipInputPause` pattern already exists on `move()` (line 1525) and is used by `temporarilyShow()` when moving items. This PR extends the same pattern to `click()`.

### DIFF
--- a/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
+++ b/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
@@ -1702,12 +1702,14 @@ extension MenuBarItemManager {
     /// - Parameters:
     ///   - item: The menu bar item to click.
     ///   - mouseButton: The mouse button to click the item with.
-    func click(item: MenuBarItem, with mouseButton: CGMouseButton) async throws {
+    func click(item: MenuBarItem, with mouseButton: CGMouseButton, skipInputPause: Bool = false) async throws {
         guard let appState else {
             throw EventError.cannotComplete
         }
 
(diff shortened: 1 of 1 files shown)


stonerl (approved):
The delay was used to make sure the icons have settled before clicked but since we made so many changes, this might not be necessary anymore. Thank you.

PR #191, opened 2026-02-23
perf: centralize 5fps image refresh loop into MenuBarItemImageCache

7a6163:
## What does this PR do?

  Centralizes the 5fps image refresh loop into `MenuBarItemImageCache`, replacing three independent per-view loops that caused duplicate Window Server screen captures when
  multiple panels were open simultaneously.

  ## PR Type

  - [x] Performance improvement
  - [x] Refactor

  - [x] No

  Three SwiftUI views (`IceBarContentView`, `MenuBarSearchContentView`, `MenuBarLayoutSettingsPane.layoutBars`) each run their own independent `while !Task.isCancelled {
  sleep(200ms); refreshImages(...) }` loop. When multiple views are visible simultaneously (e.g. IceBar + Search), this causes duplicate Window Server IPC screen captures — up
   to 15 captures/sec instead of the intended 5.

  ## What is the new behavior?

  A single centralized refresh loop in `MenuBarItemImageCache` that:

  - **Starts** when any consumer view becomes visible (`isIceBarPresented`, `isSearchPresented`, or settings layout pane)
  - **Stops** when no consumer views are visible
  - **Determines sections dynamically**: IceBar → current section only; Search/Layout → all sections
  - **Runs one capture per 200ms tick** regardless of how many views are open

  ### Files changed

  | File | Change |
  |------|--------|
  | `MenuBarItemImageCache.swift` | Added `liveRefreshTask`, `startLiveRefreshIfNeeded()`, `runLiveRefreshLoop(appState:)`, and Combine observer on navigation state |
  | `IceBar.swift` | Removed `.task { while ... refreshImages ... }` from `IceBarContentView` |
  | `MenuBarSearchPanel.swift` | Removed `.task { while ... refreshImages ... }` from `MenuBarSearchContentView` |
  | `MenuBarLayoutSettingsPane.swift` | Removed `.task { while ... refreshImages ... }` from `layoutBars` |

  ## PR Checklist

  - [ ] I've checked for console errors or crashes
  - [ ] I've run relevant tests and they pass

  ## Other information

  The Combine observer watches `isIceBarPresented`, `isSearchPresented`, `isSettingsPresented`, `settingsNavigationIdentifier`, and `isAppFrontmost` with a 50ms debounce. The
  loop uses `continue` (not `break`) when no consumer is visible mid-iteration, avoiding a race with `IceBarPanel.close()` where `currentSection` is nilled before
  `isIceBarPresented` is set to `false`.

  ## Notes for reviewers

  - The settings layout path now requires `isAppFrontmost` — an intentional optimization since the pane is invisible when Thaw is not frontmost. The old per-view loop had no
  such guard.
  - The loop does **not** self-clean `liveRefreshTask = nil` on exit to avoid a race where an old task's cleanup overwrites a newly assigned task reference. Cleanup is handled
   entirely by `startLiveRefreshIfNeeded()` via the Combine observer.
  - Verify: open IceBar → animated icons refresh; open Search → refresh; open Settings > Menu Bar Layout → refresh; open multiple simultaneously → CPU usage stays flat vs.
  before.

### DIFF
--- a/Thaw/MenuBar/IceBar/IceBar.swift
+++ b/Thaw/MenuBar/IceBar/IceBar.swift
@@ -370,21 +370,6 @@ private struct IceBarContentView: View {
             try? await Task.sleep(for: .seconds(2))
             loadingTimedOut = true
         }
-        .task {
-            // Refresh captured images at ~5fps so animated menu bar
-            // icons (e.g. Google Drive sync spinner) stay up-to-date
-            // while keeping CPU/GPU usage low.
-            while !Task.isCancelled {
-                try? await Task.sleep(for: .milliseconds(200))
-                guard !Task.isCancelled else { break }
-                let currentItems = items
-                guard !currentItems.isEmpty else { continue }
-                await imageCache.refreshImages(
-                    of: currentItems,
-                    scale: screen.backingScaleFactor
-                )
-            }
-        }
     }
 
     private static let diagLog = DiagLog(category: "IceBar.Content")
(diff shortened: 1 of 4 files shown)


stonerl (approved):
Thank you for cleaning up behind me.

PR #192, opened 2026-02-23
feat: persist and restore menu bar item order within sections

7a6163:
## Summary

  - Saves the relative order of menu bar items within each section
  (visible/hidden/always-hidden) to UserDefaults, and restores it when items are
  detected out of order — fixes apps like Stats.app whose icons get rearranged after
  restart
  - Order is saved on every cache update and restored during the cache pipeline (after
  pending relocations, before final caching)
  - Three anti-loop layers prevent restore→cache→restore cycles: `isRestoringItemOrder`
   flag suppresses saves during restore, equality check in save skips unchanged writes,
   equality check in restore skips sections already in correct order

  Closes #190 

  ## How it works

  1. **Save:** After each cache update in `uncheckedCacheItems()`,
  `saveSectionOrder(from:)` extracts per-section `uniqueIdentifier` lists and persists
  them (skips write if unchanged)
  2. **Restore:** In `cacheItemsRegardless()`, after `relocatePendingItems()`,
  `restoreSavedItemOrder()` compares current item positions against the saved order.
  For each out-of-order section, it finds a valid anchor (rightmost item, not
  temporarily shown) and moves subsequent items left-to-right using
  `.leftOfItem(anchor)`
  3. **Reset:** `resetLayoutToFreshState()` clears the saved order so no stale restore
  is attempted after a layout reset
  4. **Race condition prevention:** `isRestoringItemOrder` stays `true` through the
  post-restore recache (300ms delay), preventing intermediate/unsettled positions from
  overwriting the saved order

  ## Test plan

  - [ ] Rearrange items via Layout Bar → quit/relaunch Thaw → verify order preserved
  - [ ] Arrange Stats.app items → quit Stats → relaunch Stats → verify items return to
  saved positions
  - [ ] "Reset Layout" → verify saved order is cleared, no restore attempted
  - [ ] Launch a new menu bar app → verify it appears without disrupting existing order
  - [ ] Check diagnostic logs for "Saved section order" and "Restoring saved item
  order" messages

PR #193, opened 2026-02-23
fix: prevent infinite reorder loop in restoreSavedItemOrder

7a6163:
## Summary

  - Fixes infinite loop where `restoreSavedItemOrder` repeatedly moves items, schedules
   a recache, and the recache triggers another restore because macOS hasn't settled the
   moves within 300ms
  - Adds `guard !isRestoringItemOrder else { return false }` so the post-restore
  recache skips re-entry into the restore logic and falls through to normal caching

  Fixes https://github.com/stonerl/Thaw/pull/192#issuecomment-3943339264

  ## Root cause

  1. `restoreSavedItemOrder()` detects out-of-order items, moves them, returns `true`
  2. `isRestoringItemOrder = true` is set, recache scheduled after 300ms
  3. Recache calls `cacheItemsRegardless` → calls `restoreSavedItemOrder` again
  4. macOS hasn't fully settled the moves yet, so items still appear "out of order"
  5. Restore triggers again → schedules another recache → infinite loop

  ## Test plan

  - [ ] Rearrange items via Layout Bar → quit/relaunch Thaw → verify order restored
  without looping
  - [ ] Check diagnostic logs: "Restored saved item order; scheduling recache" should
  appear at most once per app restart, not continuously
  - [ ] Verify "Reset Layout" still clears saved order correctly

### DIFF
--- a/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
+++ b/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
@@ -2680,6 +2680,9 @@ extension MenuBarItemManager {
     ) async -> Bool {
         guard !savedSectionOrder.isEmpty else { return false }
 
+        // Don't attempt another restore while a previous restore's recache is in flight.
+        guard !isRestoringItemOrder else { return false }
+
         // Don't restore while suppressing relocations (first launch / reset).
         guard !suppressNextNewLeftmostItemRelocation else { return false }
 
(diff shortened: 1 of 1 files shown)


stonerl (approved):

stonerl:
Better, thanks. But I noticed that item reordering is broken. Either by ordering them in the menu bar directly or in the settings. They just jump back to their original position, sometimes. Have a look:

https://github.com/user-attachments/assets/c8bc2384-53d2-4f7b-a8f0-d68541945e7c

PR #197, opened 2026-02-23
fix(menubar): only restore item order when window IDs change

7a6163:
## Summary

  - Prevents `restoreSavedItemOrder` from undoing user's manual item reordering (drag
  in LayoutBar or Cmd+drag in menu bar)
  - Adds a window ID set comparison: restore only triggers when items were recreated
  (app restart → new window IDs), not when merely repositioned (user drag → same window
   IDs)

  Fixes item reordering being reverted as reported in
  https://github.com/stonerl/Thaw/pull/193#issuecomment-3943565473.

  ## Root cause

  `restoreSavedItemOrder()` runs in the cache pipeline **before** `saveSectionOrder()`.
   When a user drags items to reorder them, the next cache cycle detects the new
  positions don't match the (stale) saved order and moves everything back — causing
  items to "jump back to their original position."

  ## How the fix works

  macOS assigns new `CGWindowID` values when an app recreates its menu bar items (e.g.
  after restart). User drag-and-drop only repositions existing windows — their IDs stay
   the same.

  By comparing `Set(currentWindowIDs) != Set(previousWindowIDs)`, we precisely
  distinguish:
  - **App restart** (IDs changed) → restore saved order ✓
  - **User drag** (IDs unchanged) → skip restore, let `saveSectionOrder` persist new
  arrangement ✓

  ## Test plan

  - [ ] Reorder items via LayoutBar drag → verify order persists (not reverted)
  - [ ] Reorder items via Cmd+drag in menu bar → verify order persists
  - [ ] Quit and relaunch Stats.app → verify items restored to saved positions
  - [ ] Reset Layout → verify no restore attempted (savedSectionOrder cleared)
  - [ ] Launch a new menu bar app → verify existing item order not disrupted

### DIFF
--- a/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
+++ b/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
@@ -805,7 +805,11 @@ extension MenuBarItemManager {
                 return
             }
 
-            let didRestoreOrder = await restoreSavedItemOrder(items, controlItems: controlItems)
+            let didRestoreOrder = await restoreSavedItemOrder(
+                items,
+                controlItems: controlItems,
+                previousWindowIDs: previousWindowIDs
+            )
 
             if didRestoreOrder {
                 // Keep isRestoringItemOrder true through the recache to prevent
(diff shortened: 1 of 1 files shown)


stonerl (approved):

stonerl:
Better, but moving icons is still very erratic.

https://github.com/user-attachments/assets/fe67f389-40b0-47e3-a8d3-a2a532b1588a

PR #202, opened 2026-02-23
fix(events): include unmanaged menu bar items in bounds lookup

### LINKED ISSUE #200: [Bug]: 1.1.99-beta.3 intermittently triggers when clicking on existing menu bar apps

### Description

1.1.99-beta.3 is occasionally activating even when I click on menu bar apps, like the clock (does still work on empty part of the menu bar).

Image

Gone back to 1.1.10 (14), which works fine.

### Steps to Reproduce

MacBook Pro 13"
Intermittent (sorry!)

### App Version

1.1.99-beta.3

### macOS Version

26.3

### Monitor Setup

Single Monitor

### Additional Information

_No response_

7a6163:
## Summary

  Fixes #200

  - Commit `b546218` introduced a performance optimization that cached window bounds to
   avoid per-event Window Server IPC calls, but the cache only included **managed
  items** (items assigned to Thaw's visible/hidden/always-hidden sections)
  - Unmanaged items like Clock, Control Center, and Notification Center were excluded
  from the cache, causing clicks on them to be misidentified as clicks on empty menu
  bar space, which incorrectly triggered the hidden section
  - The fix extends `rebuildWindowBoundsLookup` to also query all on-screen menu bar
  windows via `Bridging.getMenuBarWindowList()` and include unmanaged items in the
  cache, while preserving the performance optimization (IPC calls only happen during
  cache rebuilds, not per mouse event)

  ## Test plan

  - [ ] Click on Clock — hidden section should NOT trigger
  - [ ] Click on Control Center — hidden section should NOT trigger
  - [ ] Click on Notification Center — hidden section should NOT trigger
  - [ ] Click on empty menu bar space — hidden section should still trigger as expected
  - [ ] Verify show-on-hover still works correctly
  - [ ] Test with notifications present in Notification Center
  - [ ] Test on single and multi-monitor setups

### DIFF
--- a/Thaw/Events/HIDEventManager.swift
+++ b/Thaw/Events/HIDEventManager.swift
@@ -221,13 +221,32 @@ final class HIDEventManager: ObservableObject {
     }
 
     /// Rebuilds the window bounds lookup table from the current item cache.
+    ///
+    /// Includes ALL menu bar item windows (both managed and unmanaged) so that
+    /// clicks on unmanaged items like Clock and Control Center are correctly
+    /// detected as being on a menu bar item, not on empty space.
     private func rebuildWindowBoundsLookup(from cache: MenuBarItemManager.ItemCache) {
+        // Start with managed items whose bounds are already known.
         let items = cache.managedItems
+        var knownWindowIDs = Set<CGWindowID>()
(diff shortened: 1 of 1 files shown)


stonerl (approved):

stonerl:
LGTM

PR #209, opened 2026-02-24
fix(xpc): reduce memory usage in MenuBarItemService PID cache

7a6163:
## Summary

  - Add negative caching to skip repeated AX API calls for the ~95% of running apps
  that don't have menu bar items, avoiding 200-400+ wasted calls per batch scan
  - Wrap batch scan loop in `autoreleasepool` to free AXUIElement CF objects per
  iteration instead of letting them accumulate across all apps
  - Add early termination to stop iterating apps once all menu bar windows have been
  resolved
  - Add per-`CachedApplication` `NSLock` to prevent data races between scan and cleanup
   paths
  - Remove dead code (`stableBounds(for:)`, `updatePID(for:)`)

  ## Context

  The `MenuBarItemService` XPC service was being killed by macOS after ~42 minutes due
  to excessive memory growth. XPC services have tight memory limits (typically
  50-80MB). The service resolves source PIDs for menu bar items on macOS 26+ via the
  Accessibility API, which creates CF-bridged objects on every call.

  Three factors contributed to unbounded growth:
  1. **No negative caching** — every cache miss re-queried all ~200-400 running
  processes via AX, even though only ~20-30 have menu bar items
  2. **No autoreleasepool** — `AXUIElement` CF objects created in the GCD-dispatched
  batch loop accumulated until thread pool drain
  3. **No early exit** — the loop always iterated ALL apps even after every window was
  matched

  ## Test plan

  - [ ] Build and run on macOS 26
  - [ ] Verify menu bar items are correctly identified (source PIDs resolved)
  - [ ] Monitor XPC service memory in Activity Monitor — should stabilize at a much
  lower baseline
  - [ ] Quit and relaunch a menu bar app (e.g. Stats, Bartender) to verify negative
  cache reset discovers newly-registered status items
  - [ ] Leave running for 1+ hours to confirm service is no longer killed

### DIFF
--- a/MenuBarItemService/SourcePIDCache.swift
+++ b/MenuBarItemService/SourcePIDCache.swift
@@ -103,14 +103,15 @@ final class SourcePIDCache {
             return bar
         }
 
-        /// Resets cached accessibility state so the app will be
-        /// re-queried on the next scan. Called during cleanup to:
-        /// - Release stale AXUIElement Mach port references
-        /// - Discover apps that register status items after launch
-        func resetCachedState() {
+        /// Resets the negative cache so the app will be re-checked
+        /// on the next scan. Called during cleanup to discover apps
+        /// that register status items after launch. Preserves a
+        /// valid `_extrasMenuBar` to avoid unnecessary AX re-queries.
+        func resetNegativeCache() {
             lock.withLock {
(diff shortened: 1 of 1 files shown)


stonerl (approved):
LGTM

PR #210, opened 2026-02-24
Update language flags in README.md

FormalSnake:
I updated the flags in the table, because I saw Spain and Mexico's flags being split, and noticed that English had the American flag only, same goes for dutch (in Belgium a bit more than half of us speak dutch)

You can make this change yourself too (or not) if I don't deserve to be in the contributors list for just this PR, i don't care.

Anyway, thanks for this project, because I loved Ice


stonerl (approved):
No worries. I just added the flags because I decided to the native names for the language in the list. So adding more flags is not an issue for now.

FormalSnake:
Alright, thank you for merging!

Issue #224, opened 2026-02-26
[Bug]: 1.1.99-beta.7 - icons suddenly start shuffling themselves [possibly fixed]


luckman212:
# ⚠️ Update

To anyone experiencing this bug, try [1.1.99-beta.9](https://github.com/stonerl/Thaw/releases/tag/1.1.99-beta.9) and make sure to Reset Layout after updating.

### Search for Similar Reports

- [x] I have searched existing issues for similar reports

### Description

Never saw this behavior before beta-6... 

Suddenly the menu items started re-arranging themselves, with a lot of flicker, and I couldn't move my mouse cursor (mouse cursor felt locked/stuck).

This lasted about 10-15 seconds, with icons constantly shifting positions in the menubar, sort of "fighting" for a position. Finally it stopped, and settled on a random order which was not my saved arrangement. Quitting and restarting Thaw at that point did not restore the correct layout. I had to press Reset Layout and re arrange everything. 

I tried to grab a screen recording of it, but before I could do it, it had stopped. If it happens again I will try to capture it.

### Steps to Reproduce

unknown

### App Version

- 1.1.99-beta.6 (20)
- 1.1.99-beta.7 (21)

### macOS Version

26.3

### Monitor Setup

Single Monitor

stonerl:
Install beta.7 and reset the layout.

luckman212:
Will do! Testing now 👍

luckman212:
@stonerl I'm still hitting this bug with beta-7

I was able to grab a screen recording of it!

https://github.com/user-attachments/assets/6310f86c-f43e-4378-8ec3-6141c420aefc

Here is how my layout was supposed to be set (it was jumbled after the bug hits, but I've reset it several times now and even after fixing it, after some time it eventually re-scrambles itself)

<img width="1066" height="602" alt="Image" src="https://github.com/user-attachments/assets/d4b75c9f-5d4b-4e2c-8338-37ac4997cf79" />

stonerl:
@7a6163 any idea, I cannot reproduce this for some reason.

luckman212:
If it helps at all, this never happened prior to beta-6

So it seems to be something very recent

stonerl:
Does this happen when one of the misplaced icons appears or just random?

luckman212:
99% sure there were no new icons at the time. I was just working in Sublime Text. So yeah as far as I can tell randomly happened. Will pay close attention to what I was doing next time it happens.

[...]

PR #233, opened 2026-02-27
perf(icebar): fix click latency and 5-second eventSemaphore hang

7a6163:
## Summary

  - **Fix 1 – Click latency**: `rehideTemporarilyShownItems` now uses a 50ms settle
  delay (down from 250ms) when called from `temporarilyShow`, eliminating the
  noticeable pause between consecutive IceBar clicks.

  - **Fix 2 – Adaptive timeout floor**: Raised the minimum move-operation timeout from
  25ms to 75ms, preventing the adaptive algorithm from compressing the timeout below
  the polling interval of `waitForMoveEventResponse` (10ms × ~7 polls) and causing
  unnecessary retry cascades.

  - **Fix 3 – eventSemaphore 5-second hang**: `postEventWithBarrier` and
  `scrombleEvent` used `withCheckedThrowingContinuation` inside an unstructured inner
  `Task`. The inner task's body (`enable()` + `post()`) completes almost immediately;
  by the time `Task(timeout:)` fires, cancelling it is a no-op and the continuation is
  stuck forever, holding `eventSemaphore` for 5 seconds. Fixed by storing the
  continuation in an outer-scope `OSAllocatedUnfairLock` so the outer
  `withTaskCancellationHandler.onCancel` can resume it directly. All three resume paths
   (EventTap1, innerTask onCancel, outer onCancel) are guarded by
  `didResume.tryClaimOnce()` to guarantee exactly-once resumption.

  - **Semaphore consistency**: `postClickEvents` now uses `Task.detached` to signal
  `eventSemaphore`, matching `postMoveEvents` and avoiding an unnecessary MainActor
  scheduling hop.

  - **Dead code removal**: Removed duplicate `Duration.clamped(min:max:)` extension;
  the existing generic `Comparable.clamped` in `Extensions.swift` already covers
  `Duration`.

  - **Composite capture improvement**: Pre-filter off-screen items using a single
  `getWindowBounds` call before composite capture, eliminating width-mismatch failures
  that caused full cache invalidation on each cycle.

  ## Root Cause (Fix 3 detail)

  Task(timeout:) uses withThrowingTaskGroup internally.
  withThrowingTaskGroup MUST await ALL child tasks before returning.

  Timeline:
    t=0ms    innerTask starts → enable() + post() → completes immediately
    t=Xms    timeout fires → outer task cancelled → onCancel called
    t=Xms    innerTask.cancel() → no-op (already done)
    t=Xms    withCheckedThrowingContinuation stuck (no one to resume it)
    t=Xms    withThrowingTaskGroup waits forever
    t=X+5000ms  eventSemaphore.wait() times out → forced signal

  Fix: outer `onCancel` now directly resumes the continuation via `continuationHolder`.

  ## Test Plan

  - [ ] Open IceBar and click multiple items in succession — response should be
  immediate with no 250ms pause between clicks
  - [ ] Trigger move operations for apps previously causing `eventSemaphore timed out`
  (e.g., PasteNow, Telegram) — should no longer see 5-second hangs in logs
  - [ ] Verify adaptive timeout stabilises above 75ms and stops compressing toward zero
  - [ ] Enable diagnostic logging (Advanced Settings) and confirm no `eventSemaphore
  timed out` log entries after several minutes of use
  - [ ] Confirm composite capture no longer logs width-mismatch errors on standard
  display setups

### DIFF
--- a/Thaw/MenuBar/IceBar/IceBar.swift
+++ b/Thaw/MenuBar/IceBar/IceBar.swift
@@ -524,13 +524,19 @@ private struct IceBarItemView: View {
             guard let itemManager, let menuBarManager else {
                 return
             }
+            let clickStartTime = Date.now
+            IceBarItemView.diagLog.debug("leftClick: user clicked \(item.logString)")
             menuBarManager.section(withName: section)?.hide()
             Task {
                 try await Task.sleep(for: .milliseconds(25))
                 if Bridging.isWindowOnScreen(item.windowID) {
                     try await itemManager.click(item: item, with: .left)
+                    let duration = Date.now.timeIntervalSince(clickStartTime)
+                    IceBarItemView.diagLog.debug("leftClick: ✓ completed in \(Int(duration * 1000))ms (on-screen path)")
                 } else {
                     await itemManager.temporarilyShow(item: item, clickingWith: .left, on: displayID)
+                    let duration = Date.now.timeIntervalSince(clickStartTime)
+                    IceBarItemView.diagLog.debug("leftClick: ✓ completed in \(Int(duration * 1000))ms (temp-show path)")
                 }
             }
         }
--- a/Thaw/MenuBar/MenuBarItems/MenuBarItem.swift
+++ b/Thaw/MenuBar/MenuBarItems/MenuBarItem.swift
@@ -376,8 +376,14 @@ extension MenuBarItem {
             }
         }
 
-        let nilPIDCount = items.filter { $0.sourcePID == nil }.count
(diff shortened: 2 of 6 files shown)


7a6163:
@stonerl Could you test this version first?

stonerl:
I'll do so. I let you know my findings.

stonerl (approved):

stonerl:
LGTM. I tested a couple of hours now and the random icon moving also seems to be solved. Great work, thanks allot.

stonerl:
Okay no it did not solve the shuffling of icons, reported in #224, but that also was not addressed in this PR.

PR #236, opened 2026-02-28
Update Indonesian translation

volcbs:
Adding new strings for Indonesian translation.


stonerl (approved):
Nice, thank you very much.

Issue #62, opened 2026-02-11
[Bug]: OneDrive - if 2 accounts, one disappears if hidden


ksignorini:
### Search for Similar Reports

- [x] I have searched existing issues for similar reports

### Description

I have 2 OneDrive icons in my menu bar. One is work. One is personal. Without Thaw running, I see both. With Thaw running, Settings only shows me one icon that I can put into the hidden section; yet I still see one in hidden and one in not hidden in the actual menu bar. Cmd-dragging the not hidden one to the hidden area makes that one disappear entirely.

### Steps to Reproduce

See above.

### App Version

1.0.0 and 1.0.1 beta

### macOS Version

26.2

### Monitor Setup

Multi-monitor Setup

### Additional Information

Laptop in clamshell with single Studio Display connected.

stonerl:
Thank you for reporting I need to check how OneDrive reports a second icon to the system. Could you enable and attach die diagnostic log?

stonerl:
Looks like Microsoft decided to use the same identifier for both icons

```
uncheckedCacheItems: skipping duplicate tag <com.microsoft.OneDrive-mac:Item-1 (windowID: 3848)>
```

emaba:
I know that it's hard to fix, but it's really annoying.
Having the same identifier, are there any other ways to identify them?

emaba:
I agree with this comment:

https://github.com/jordanbaird/Ice/issues/857#issuecomment-3902043957

vaultmaster:
Actually, it would be great if you can somehow identify which process/account is attached to the icon and then replace the icon with a unique one, e.g. with an overlay such as P for personal and B for business, that would fix the OneDrive mess :)

stonerl:
I see what I can do 😁

stonerl:
This will be available with the next development release and will need some thorough testing. I'll let you know when it is available, in the meantime you can watch this video.

https://github.com/user-attachments/assets/d232820e-43a3-44f8-bbff-5e3b46254737

vaultmaster:
Now that is a response, looking forward to testing it!

stonerl:
Beta is out for testing: 1.1.99-beta.1

vaultmaster:
Appears to work as expected at first glance, will test a bit more!

vaultmaster:
Then again, it moved back to 1 icon in the menu bar, if I select it, I get the 'other' one.

stonerl:
You mean it opens the wrong instance/account?

[...]

PR #241, opened 2026-03-02
fix(menubar): distinguish items with same icon from different apps

Tsingv:
When two apps register menu bar items with the same icon (and thus the same namespace + title on macOS 26), Thaw could conflate them into a single logical item. This caused:
- Clicking one item could activate the wrong one
- Auto-hide (rehide) would twitch and leave one item stuck visible
- Pending relocation entries collided, preventing correct rehide

image

Root causes and fixes:

1. `matchesIgnoringWindowID` ignored both windowID and instanceIndex. Two items distinguished only by instanceIndex were treated as one in click handling, bounds lookup, temporarily-shown matching, and image cache cleanup. Now compares instanceIndex as well.

2. PID propagation on macOS 26 could assign App A's PID to App B's unresolved item when both shared a generic title like "Item-0". Now only propagates when the resolved PID already owns 2+ items, confirming it is a multi-item app.

3. Runtime identifiers for pendingRelocations, pendingReturnDestinations, and activelyShownTags used "namespace:title" without instanceIndex, causing key collisions. Introduced `MenuBarItemTag.tagIdentifier` which appends instanceIndex when nonzero, and adopted it across all temporarily-shown and section-restore code paths.

### DIFF
--- a/Thaw/MenuBar/MenuBarItems/MenuBarItem.swift
+++ b/Thaw/MenuBar/MenuBarItems/MenuBarItem.swift
@@ -322,12 +322,24 @@ extension MenuBarItem {
         // may fail for one of the windows due to timing skew between
         // CG and AX coordinate updates.
         //
-        // If an unresolved item shares the same window title with a
-        // resolved item, it almost certainly belongs to the same app.
-        // Re-create the item with the sibling's sourcePID so it gets
-        // the correct namespace, display name, and app icon.
+        // Only propagate a resolved PID to unresolved items sharing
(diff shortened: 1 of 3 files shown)


Tsingv:
Oh, I found a similar issue on [#62](https://github.com/stonerl/Thaw/issues/62). It's might be a alternative way to solve the problem.

Tsingv:
image
Also, it works fine when reorder the icons.

stonerl (approved):
LGTM Thank you very much for your contribution.

PR #244, opened 2026-03-02
fix(menubar): suppress restore during startup settling period to prevent icon parade

### LINKED ISSUE #224: [Bug]: 1.1.99-beta.7 - icons suddenly start shuffling themselves [possibly fixed]

7a6163:
## Problem

Addresses the \"icon parade\" reported in #224 — icons continuously shuffling for 10–15 seconds. The root cause is that `restoreItemsToSavedSections` and `restoreSavedItemOrder` fire on **every** `didLaunchApplicationNotification` (1 s debounce). During login, 20+ apps load over ~30 seconds, each triggering its own restore cycle that conflicts with the next, producing cascading icon moves.

A secondary trigger is app update checks (e.g. BetterTouchTool), which briefly remove and re-add a menu bar icon, causing the same back-to-back restore cascade outside of login.

### Startup settling period

Introduce `isInStartupSettling`, a flag that blocks restore operations and `saveSectionOrder()` calls for a fixed window after `performSetup()` completes:

- **30 s** when `ProcessInfo.processInfo.systemUptime < 60` (login item / auto-launch boot)
- **5 s** otherwise (manual restart, app-update scenario)

During the period, every `didLaunchApplicationNotification` still triggers a cache cycle so the UI stays accurate, but neither `restoreItemsToSavedSections` nor `restoreSavedItemOrder` runs.

`saveSectionOrder()` is also suppressed during settling to protect the UserDefaults-loaded reference order from being overwritten by transient physical positions before restore has run.

When the period expires, one final `cacheItemsRegardless(skipRecentMoveCheck: true)` enforces the user's saved layout against whatever macOS placed items. `skipRecentMoveCheck: true` prevents the restore from being silently skipped by the 5 s move cooldown in case `relocateNewLeftmostItems` stamped a timestamp near the end of the window.

### Re-entrancy safety

`startupSettlingTask: Task<Void, Never>?` tracks the in-flight task. A subsequent `performSetup()` call cancels the previous task before starting a new one. The cancelled task uses `do/catch` on `Task.sleep` to exit cleanly without touching shared state.

A `settlingDeadline: ContinuousClock.Instant?` is stored so that re-entry computes `max(existingDeadline, newDeadline)`, preventing a second call from truncating the original 30 s login window if `systemUptime` has crossed 60 s by then.

### Layout reset interaction

`resetLayoutToFreshState()` cancels the settling task and clears the flag immediately — a user-initiated reset is authoritative and must not be blocked by an in-flight settling period.

### Move cooldown bump

Both `restoreItemsToSavedSections` and `restoreSavedItemOrder` raise their move-operation cooldown from 2 s to 5 s, providing more separation between consecutive automatic restores when apps restart in quick succession.

## Changes

Single file: `Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift` (+89 / -5)

## Test plan

- [ ] **Login item boot** — enable Launch at Login, reboot, verify no icon shuffle during the first ~30 s after login
- [ ] **App update trigger** — use BetterTouchTool's "Check for updates" and verify icons do not shuffle
- [ ] **Normal app restart** — quit and relaunch an app that has a hidden menu bar item; verify it is restored to the correct section within ~5–8 s
- [ ] **Reset Layout during boot** — open Settings and press Reset Layout within 30 s of login; verify the reset takes effect immediately without waiting for settling to expire
- [ ] **Permission re-grant** — revoke and re-grant Screen Recording permission during the settling window; verify settling deadline is preserved (not truncated)
- [ ] **Show/hide** — verify IceBar show/hide behaviour is unaffected during and after the settling period

> ⚠️ **More testing needed.** The 30 s / 5 s thresholds and the `systemUptime < 60` boundary are heuristics. Feedback on edge cases — especially on slower machines or unusual login configurations — would be very helpful before merging.

Fixes #224

### DIFF
--- a/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
+++ b/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
@@ -147,6 +147,12 @@ final class MenuBarItemManager: ObservableObject {
     private var isRestoringItemOrder = false
     /// Timestamp when isRestoringItemOrder was set (for timeout detection).
     private var isRestoringItemOrderTimestamp: Date?
+    /// True during the startup settling period, during which restore operations
+    /// and section-order saves are suppressed. This prevents cascading icon moves
+    /// when many apps launch at login (login item boot) or restart in quick succession
+    /// (e.g. app update checks). Cleared after a fixed delay, then one final
+    /// restore runs to enforce the user's saved layout.
+    private var isInStartupSettling = false
     /// Persisted bundle identifiers explicitly placed in hidden section.
     private var pinnedHiddenBundleIDs = Set<String>()
     /// Persisted bundle identifiers explicitly placed in always-hidden section.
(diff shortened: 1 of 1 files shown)


diazdesandi (commented):
Tested this and the issue persisted. macOS 24.4 Beta 3

Edit: Seems related to Beta 3.

stonerl (approved):
LGTM

PR #256, opened 2026-03-10
[codex] Align control item hover background with native menu bar items

### LINKED ISSUE #252: [Bug]: Thaw separator item looks inconsistent with native macOS menu bar items

### Description

The Thaw separator / overflow item (...) does not match the appearance and behavior of native macOS menu bar items.

Native macOS menu bar items have a consistent rounded height and a uniform hover / active background shape. In comparison, the Thaw separator item looks visually off. Its highlighted background appears to have a different height and overall sizing, which makes it stand out as inconsistent next to the surrounding system items.

This is especially noticeable when the separator item is shown beside items like the battery indicator, where native items keep a consistent capsule-like hover / active state, but the Thaw item does not align visually with them.

### Steps to Reproduce

1. Enable Thaw in the macOS menu bar
2. Click on the Thaw separator and observe active states
3. Compare the Thaw separator / overflow item (...) with nearby native menu bar items
4. Notice that the Thaw item’s background height / shape looks inconsistent with the native macOS styling

### App Version

1.1.99-beta.8

### macOS Version

26.3.1

### Monitor Setup

Single Monitor

### Additional Information

Image

Image

exsesx:
Closes #252.

The control item used for Thaw's separator and overflow affordance was constrained to the top and bottom edges of the status item content view. On current macOS builds that causes its highlighted background to occupy a different vertical footprint than adjacent native menu bar items, so the separator looks taller and visually out of place when hovered or pressed.

The underlying issue is that the button was being stretched to fill the container rather than aligned the way native status bar controls are. That makes the active and hover background shape follow the full content view height instead of the standard centered capsule geometry users expect in the menu bar.

This change replaces the top and bottom pinning with a vertical centering constraint on the status item button. Keeping the button centered lets AppKit render the control at the same apparent vertical alignment as neighboring system items, which makes the hover and active background match native menu bar behavior more closely.

I validated the change with the same checks used by the repository CI:

- `swiftlint --strict`
- `xcodebuild -project Thaw.xcodeproj -scheme Thaw -destination platform=macOS -configuration Release MACOSX_DEPLOYMENT_TARGET=14.0 CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO build`

### DIFF
--- a/Thaw/MenuBar/ControlItem/ControlItem.swift
+++ b/Thaw/MenuBar/ControlItem/ControlItem.swift
@@ -83,8 +83,7 @@ final class ControlItem {
                     }
 
                     NSLayoutConstraint.activate([
-                        button.topAnchor.constraint(equalTo: contentView.topAnchor),
-                        button.bottomAnchor.constraint(equalTo: contentView.bottomAnchor),
+                        button.centerYAnchor.constraint(equalTo: contentView.centerYAnchor),
                     ])
                 } else {
                     self.constraint = nil


stonerl (approved):
LGTM, thank you for your contribution.

PR #301, opened 2026-03-18
fix: restore item order for single-icon items when multi-icon apps present

### LINKED ISSUE #300: [Bug]: Menu bar icon order not preserved when multi-icon apps (Stats, iStat Menus) are present
## Description

When a multi-icon menu bar app is running (e.g. [Stats](https://github.com/exelban/stats), iStat Menus, MenuMeters), Thaw fails to restore saved icon order for **all** items — including single-icon apps that could be safely repositioned.

## Steps to Reproduce

1. Install a multi-icon menu bar app like [Stats](https://github.com/exelban/stats) (which registers multiple menu bar extras from a single process)
2. Arrange menu bar icons in desired order
3. Restart the Mac or log out and back in
4. Observe that menu bar icons are in a different (randomised) order

## Expected Behavior

Single-icon menu bar items should be restored to their saved positions. Multi-icon app items can be left to macOS since they self-position.

## Actual Behavior

`restoreSavedItemOrder()` bails out entirely when **any** multi-icon or indexed-item app is detected, skipping restoration for all items. Meanwhile, `saveSectionOrder()` continues to overwrite the good saved state with the randomised positions, causing the saved order to degrade over time.

## Root Cause

In `MenuBarItemManager.swift`, `restoreSavedItemOrder()` has an early return:

```swift
let hasMultiIconApps = itemsPerNamespace.values.contains { $0 > 1 }
guard !hasIndexedItems && !hasMultiIconApps else {
    return false
}
```

This treats the presence of multi-icon apps as a reason to skip **all** order restoration, rather than just excluding those specific items.

## Environment

- macOS 15 Sequoia
- Thaw 1.1.99-beta.11
- Stats (or any multi-icon menu bar app)

coygeek:
## Summary

Fixes #300

- `restoreSavedItemOrder()` now excludes multi-icon/indexed app items from ordering instead of bailing out entirely, so single-icon items are still restored to their saved positions
- Adds `restoreWasPartiallySkipped` flag to suppress `saveSectionOrder()` after a partial restore, preventing good saved state from being overwritten with randomised positions
- Clears the partial-skip flag on explicit user drag (cmd+drag) so normal save behavior resumes

## Context

Apps like [Stats](https://github.com/exelban/stats), iStat Menus, and MenuMeters register multiple menu bar icons from a single process. The previous logic detected this and skipped order restoration for *all* items, causing icons to jump around on every restart. Over time, `saveSectionOrder()` would overwrite the good saved state with randomised positions, making the problem progressively worse.

## Test plan

- [ ] Install Stats (or another multi-icon menu bar app) alongside several single-icon menu bar apps
- [ ] Arrange icons in desired order, restart Mac
- [ ] Verify single-icon items restore to saved positions
- [ ] Verify multi-icon app items are left alone (no shuffling)
- [ ] Cmd+drag an icon to a new position, restart — verify the new position is saved
- [ ] Test without any multi-icon apps installed — verify full restore still works

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented partial restores involving multi-icon or indexed items from overwriting previously saved section layouts.
  * After a partial restore, normal saving of section order is suppressed until the user performs a manual rearrangement or an explicit full restore.

* **Documentation**
  * Clarified that single-icon apps restore saved order; multi-icon apps from the same process are treated as grouped blocks and won’t block other items from being restored.

### DIFF
--- a/FREQUENT_ISSUES.md
+++ b/FREQUENT_ISSUES.md
@@ -24,7 +24,7 @@ the always-hidden section, then Command + drag the item into a different section
 
 ## Thaw does not remember the order of items
 
-This is not a bug, but a missing feature. It is being tracked in [#26](https://github.com/jordanbaird/Thaw/issues/26).
+Order restoration works for single-icon apps. Apps that register multiple menu bar icons from the same process (e.g. iStat Menus, Stats, MenuMeters) are treated as opaque blocks — their internal ordering is left to macOS, but they no longer prevent other items from being restored to their saved positions.
 
 ## How do I solve the `Thaw cannot arrange menu bar items in automatically hidden menu bars` error?
 
--- a/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
+++ b/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
@@ -147,6 +147,11 @@ final class MenuBarItemManager: ObservableObject {
     private var isRestoringItemOrder = false
     /// Timestamp when isRestoringItemOrder was set (for timeout detection).
     private var isRestoringItemOrderTimestamp: Date?
+    /// Indicates the most recent order restore was partial (multi-icon/indexed
+    /// items were present). While true, `saveSectionOrder` is suppressed to
+    /// prevent overwriting good saved state with potentially randomised
+    /// positions. Cleared on explicit user drag or a full (non-partial) restore.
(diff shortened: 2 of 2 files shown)


coygeek:
Thank you for the 1.2.0-rc.1, which fixes this issue.

PR #338, opened 2026-03-28
fix: reduce IceBar click jitter with fast-path temporarilyShow

CamilleGuillory:
## Problem

When clicking hidden items from the IceBar, the icon visibly jitters/shifts in the menu bar for ~1.8 seconds before the menu opens. This is caused by:

1. `move()` retrying 8 times with exponentially increasing timeouts, even though the first attempt always repositions the item close enough (position verification fails due to slight drift after mouseUp)
2. ~300ms of post-move delays (250ms settle wait, item re-fetch, 25ms extra sleep) before the click fires

## Fix

Add a `fastPath` mode for IceBar clicks that:
- Limits `move()` to 1 attempt (via new `maxMoveAttempts` parameter, default 8 for existing callers)
- Skips the settle wait, item re-fetch, and extra sleep between move and click

This reduces the visible jitter from ~1.8s to a brief flash (~100ms).

## Changes

- `MenuBarItemManager.swift`: Add `maxMoveAttempts` parameter to `move()` and `fastPath` parameter to `temporarilyShow()`
- `IceBar.swift`: Pass `fastPath: true` for both left and right click handlers

## Testing

Tested on macOS Tahoe with 30+ menu bar items across two displays. Hidden items clicked from the IceBar now open their menus with minimal visible jitter. No regression for non-IceBar move operations (layout, rehide, etc.) which continue using the default 8-attempt path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Faster interactions for off-screen menu bar items — clicks register more quickly with reduced delays.
  * Fewer move retries when repositioning items, improving reliability and reducing repeated attempts.
  * Improved reveal timing for temporary menu bar item shows to reduce missed or duplicate clicks.

### DIFF
--- a/Thaw/MenuBar/IceBar/IceBar.swift
+++ b/Thaw/MenuBar/IceBar/IceBar.swift
@@ -539,7 +539,7 @@ private struct IceBarItemView: View {
                     let duration = Date.now.timeIntervalSince(clickStartTime)
                     IceBarItemView.diagLog.debug("leftClick: ✓ completed in \(Int(duration * 1000))ms (on-screen path)")
                 } else {
-                    await itemManager.temporarilyShow(item: item, clickingWith: .left, on: displayID)
+                    await itemManager.temporarilyShow(item: item, clickingWith: .left, on: displayID, fastPath: true)
                     let duration = Date.now.timeIntervalSince(clickStartTime)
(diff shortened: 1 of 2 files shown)


stonerl:
Thank you. And sorry for the long wait.

PR #305, opened 2026-03-19
[Feat] Add optional setting for using Option-clicking to open Always Hidden section

gitmichaelqiu:
## What does this PR do?

Implement a setting (optional, default off) to open Always Hidden bar upon Option+clicking and hides the bar after double clicking, behaving similar to Ice.

## PR Type

- [ ] Bugfix
- [ ] CI/CD related changes
- [ ] Code style update (formatting, renaming)
- [ ] Documentation
- [x] Feature
- [ ] i18n/l10n (Translation/Localization)
- [ ] Refactor
- [ ] Performance improvement
- [ ] Test addition or update
- [ ] Other (please describe):

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## What is the current behavior?

Thaw currently has a different Always Hidden bar opening logic to Ice.

- In Thaw, click twice opens the Always Hidden bar; In Ice, it hides the bar
- In Ice, Option + click opens the Always Hidden bar

Issue Number: N/A

## What is the new behavior?

A new setting called "Use Option-click to open always hidden section" is added. When this is off, the bar behaves the same as before. If this is on, the bar will hide after a double-click; the Always Hidden bar will show when Option-clicked.

This would make the app more friendly to those just transitioned here from Ice.

Localization for this line is also added.

## PR Checklist

- [x] I’ve built and run the app locally
- [x] I’ve checked for console errors or crashes
- [x] I've run relevant tests and they pass
- [x] I've added or updated tests (if applicable)
- [x] I've updated documentation as needed
- [x] I've verified localized strings work correctly (if i18n/l10n changes)

## Other information

Screenshot 2026-03-19 at 18 15 34

The new setting is added in Advanced > Menu Bar Sections.

## Summary by CodeRabbit

* **New Features**
  * Added a Settings toggle: "Use Option‑click to open always‑hidden section."

* **Behavior Changes**
  * Option‑click now toggles the always‑hidden menu section when enabled; when disabled, double‑click/standard interactions maintain previous visibility behavior.
  * Click handling now explicitly respects the new setting.

* **Bug Fixes**
  * Menu section hiding logic simplified for more consistent hide/show behavior across section types.

PR #385, opened 2026-04-16
[Fix] #305 More robust Option + Click detection

gitmichaelqiu:
## What does this PR do?

I found that Option + Click was not working after #305 was merged into `development` branch (it worked when I opened the PR though. I recently rebuilt the app from the source code finding it not working). The issue was caused by that I used `modifierFlags == .option` condition to detect Option + Click, which could be too strict, as macOS mouse events often include additional system-level bits.

## PR Type

- [x] Bugfix
- [ ] CI/CD related changes
- [ ] Documentation
- [ ] Feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] Test addition or update

- [ ] Yes
- [x] No

Option + Click will only open the hidden bar (after #305 was merged).

Issue Number: #305

## What is the new behavior?

Simply fixed the error by using `intersection(.deviceIndependentFlagsMask) == .option`, which is a more robust method.

No changes are brought to the case when the setting for Option + Click is off.

## PR Checklist

- [x] I've verified localized strings work correctly (if i18n/l10n changes)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced modifier key detection for Control and Option clicks for improved reliability.

* **Refactor**
  * Updated menu bar section visibility behavior based on advanced settings configuration.

### DIFF
--- a/Thaw/MenuBar/ControlItem/ControlItem.swift
+++ b/Thaw/MenuBar/ControlItem/ControlItem.swift
@@ -577,15 +577,13 @@ final class ControlItem {
                     return
                 }
 
-                if modifierFlags == .control {
+                if modifierFlags.intersection(.deviceIndependentFlagsMask) == .control {
                     showMenu()
                     return
                 }
 
-                if modifierFlags == .option {
-                    // Option-click: only toggle always-hidden if enabled.
+                if modifierFlags.intersection(.deviceIndependentFlagsMask) == .option {
                     if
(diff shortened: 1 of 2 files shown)


stonerl:
@gitmichaelqiu Thank you. I think I fixed this already with this commit: 2b0c99b

gitmichaelqiu:
Oh didn't notice it. I will close the issue.

stonerl:
No worries, we were probably working on this fix at the exact same time :D

PR #420, opened 2026-04-20
refactor: migrate captureScreenBelowWindow to ScreenCaptureKit

stonerl:
## What does this PR do?

Replace deprecated CGWindowListCreateImage API with modern ScreenCaptureKit:

- Use SCShareableContent to enumerate displays and windows
- Use SCContentFilter to capture display while excluding menu bar window
- Use SCStream for async single-frame capture
- Remove old sync function (macOS 14+ requirement)

Updates MenuBarOverlayPanel to use new async method with existing
UpdateTaskContext infrastructure.

## PR Type

- [ ] Bugfix
- [ ] CI/CD related changes
- [ ] Code style update (formatting, renaming)
- [ ] Documentation
- [ ] Feature
- [x] i18n/l10n (Translation/Localization)
- [x] Refactor
- [x] Performance improvement
- [ ] Test addition or update
- [ ] Other (please describe):

## Does this PR introduce a breaking change?

- [x] Yes
- [ ] No

If yes, please describe the impact and migration path:

Only macOS 26 is supported now, macOS 14/15 have been sunsetted.

## What is the current behavior?

We currently use a deprecated macOS 14 API for getting the menubar icons.

## What is the new behavior?

 This PR uses an internal macOS API `SkyLight`.

## PR Checklist

- [x] I’ve built and run the app locally
- [x] I’ve checked for console errors or crashes
- [x] I've run relevant tests and they pass
- [ ] I've added or updated tests (if applicable)
- [ ] I've updated documentation as needed
- [ ] I've verified localized strings work correctly (if i18n/l10n changes)

## Summary by CodeRabbit

* **New Features**
  * New window/screen capture pipeline with a single permission prompt; Settings adds Back/Forward controls.

* **Visual Improvements**
  * Expanded glass-effect styling across menus, panels, tooltips, and buttons.
  * Adaptive menu bar coloring and notch indicator that respond to screen brightness.

* **UI Updates**
  * Simplified, unified layouts with consistent rounded-corner styling across Settings and menu bar.
  * Refined search panel, badges, and control visuals for more consistent appearance and behavior.

PR #421, opened 2026-04-21
Prevent double-click from being stolen by newly revealed hidden items

Skyearn:
## Summary

This PR fixes a race in `Show on click` when `Double-click for always-hidden` is enabled.

Previously, the first click could reveal hidden items immediately, and the second click of a double-click would often land on one of those newly revealed items instead of being treated as the second click on empty menu bar space.

## What changed

- keep the current single-click behavior: the first click still reveals hidden items immediately
- add a temporary protected region centered on the first click location
- keep that protected region active for the system `NSEvent.doubleClickInterval`
- swallow left-clicks inside that region during the double-click window so newly revealed items cannot steal the second click
- if the second click is a double-click, show the always-hidden section
- automatically remove the protection once the double-click window expires

## Why

The previous behavior made double-click unreliable because the UI changed under the cursor after the first click.  
That meant the second click could hit a newly revealed menu bar item instead of continuing the original click sequence.

Using a first-click protection region avoids relying on menu bar item positions that may still be updating right after reveal, and keeps the second click associated with the original interaction.

## Result

- single-click still feels immediate
- the second click is no longer stolen by newly revealed hidden items
- double-click to show always-hidden is more reliable
- the implementation no longer depends on stale hidden-item bounds for guarding the second click

## Testing

Tested locally by building a signed Debug app and verifying that:

- a single click still reveals hidden items immediately
- the area around the first click is temporarily protected during the system double-click interval
- a double-click no longer falls through to newly revealed hidden items
- always-hidden can still be revealed by double-click

## Note

While working on this, I also noticed a separate startup-time issue: both `Show on hover` and `Show on click` can feel noticeably slow right after launch.

I have not included a fix for that in this PR, since the cause seems more involved and likely unrelated to the double-click guard itself. My current impression is that the startup path around input handling, menu bar state, and initial item/setup work is more complex than this PR should try to address.

I plan to investigate that separately and will see if I can narrow it down into a smaller, focused follow-up fix.

09:35

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevent hover show/hide races with tokenized delayed tasks and add a click-guard to avoid accidental single-click reveals.

* **Performance / Startup**
  * Non-blocking, multi-pass item warm‑up and skipped initial frontmost-app trigger to avoid startup work.

* **Behavior**
  * More accurate menu-bar hit-testing and bounds rebuilds; stable ordering for duplicated items and optional faster item discovery that can skip process resolution.

* **Reliability**
  * IPC moved to async/await with cancellation-safe reply handling.

* **Chores**
  * Timeout helper accepts optional parameters for simpler calls.

### DIFF
--- a/Thaw/Events/HIDEventManager.swift
+++ b/Thaw/Events/HIDEventManager.swift
@@ -695,55 +695,58 @@ extension HIDEventManager {
             return
         }
 
-        Task {
-            if isDoubleClick {
-                guard
-                    appState.settings.general.showOnClick,
-                    appState.settings.general.showOnDoubleClick
-                else {
-                    return
-                }
+        if isDoubleClick {
+            guard
+                appState.settings.general.showOnClick,
+                appState.settings.general.showOnDoubleClick
+            else {
+                return
+            }
+            Task {
                 if let alwaysHiddenSection = appState.menuBarManager.section(withName: .alwaysHidden),
                    alwaysHiddenSection.isEnabled
                 {
                     alwaysHiddenSection.show()
                 }
-            } else {
-                guard appState.settings.general.showOnClick else {
-                    return
-                }
+            }
+        } else {
+            guard appState.settings.general.showOnClick else {
+                return
+            }
 
-                if NSEvent.modifierFlags == .control {
-                    handleSecondaryContextMenu(appState: appState, screen: screen)
-                    return
-                }
+            if NSEvent.modifierFlags == .control {
+                handleSecondaryContextMenu(appState: appState, screen: screen)
+                return
+            }
 
-                if NSEvent.modifierFlags == .option {
-                    if appState.settings.advanced.useOptionClickToShowAlwaysHiddenSection,
-                       let alwaysHiddenSection = appState.menuBarManager.section(withName: .alwaysHidden),
-                       alwaysHiddenSection.isEnabled
-                    {
-                        alwaysHiddenSection.show()
-                    }
-                    return
-                }
-
-                if let hiddenSection = appState.menuBarManager.section(withName: .hidden),
-                   hiddenSection.isEnabled
+            if NSEvent.modifierFlags == .option {
+                if appState.settings.advanced.useOptionClickToShowAlwaysHiddenSection,
+                   let alwaysHiddenSection = appState.menuBarManager.section(withName: .alwaysHidden),
+                   alwaysHiddenSection.isEnabled
                 {
-                    let shouldArmGuard =
-                        appState.settings.general.showOnDoubleClick
-                            && hiddenSection.isHidden
-                            && (appState.menuBarManager.section(withName: .alwaysHidden)?.isEnabled ?? false)
-
-                    hiddenSection.toggle()
-
-                    if shouldArmGuard {
-                        armShowOnClickGuard(screen: screen, at: clickLocation)
-                    } else {
-                        disarmShowOnClickGuard()
-                    }
+                    Task { alwaysHiddenSection.show() }
                 }
+                return
+            }
+
+            if let hiddenSection = appState.menuBarManager.section(withName: .hidden),
+               hiddenSection.isEnabled
+            {
+                let shouldArmGuard =
+                    appState.settings.general.showOnDoubleClick
+                        && hiddenSection.isHidden
+                        && (appState.menuBarManager.section(withName: .alwaysHidden)?.isEnabled ?? false)
+
+                // Arm the guard synchronously before toggling so the CGEventTap
+                // is active before any second click can arrive; a Task hop would
+                // leave a window where the tap is not yet started.
+                if shouldArmGuard {
+                    armShowOnClickGuard(screen: screen, at: clickLocation)
+                } else {
+                    disarmShowOnClickGuard()
+                }
+
+                Task { hiddenSection.toggle() }
             }
         }
     }
--- a/Thaw/Events/HIDEventManager.swift
+++ b/Thaw/Events/HIDEventManager.swift
@@ -428,6 +428,11 @@ final class HIDEventManager: ObservableObject {
         var c = Set<AnyCancellable>()
 
         if let appState {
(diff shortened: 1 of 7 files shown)


Skyearn:
## What does this PR do?

This PR optimizes the startup performance of the Thaw UI library by refining the initialization logic for components that use `showonhover` and `showonclick` triggers (such as Popover, Tooltip, and Dropdown). 

The core improvement involves **lazy-mounting the overlay content**, ensuring that hidden elements are not eagerly rendered into the DOM during the initial application bootstrap.

## PR Type

- [ ] Bugfix
- [ ] CI/CD related changes
- [ ] Documentation
- [ ] Feature
- [ ] Refactor
- [x] Performance improvement
- [ ] Test addition or update

- [ ] Yes
- [x] No

Previously, components with `showonhover` or `showonclick` behaviors would often render their hidden overlay content into the DOM immediately upon startup. For pages with a large number of these components (e.g., a long list of tooltips or menus), this resulted in:
1. Increased initial DOM depth and memory usage.
2. Slower WASM instantiation and hydration times in Leptos.
3. Unnecessary CPU cycles spent on calculating styles for hidden elements.

## What is the new behavior?

- **Lazy Rendering**: Overlay content is now only rendered and mounted to the DOM when the trigger event (hover or click) actually occurs for the first time.
- **Improved Hydration**: Reduced the number of reactive nodes that Leptos needs to track during the initial hydration phase.
- **Resource Efficiency**: Significantly lowered the initial memory footprint and improved the Time to Interactive (TTI) for heavy UI pages.

## PR Checklist

## Other information

N/A

### Notes for reviewers

Please verify if there are any side effects when these components are used inside a `Transition` or `Suspense` block, especially regarding the timing of the first-time hover/click interaction.

Skyearn:
## What does this PR do?

This PR improves the reliability of hover- and click-triggered menu bar interactions, especially after repeatedly toggling `Show on hover`.

It hardens several async task/state edges in `HIDEventManager`, preserves `sourcePID` semantics for duplicated menu bar items, makes the initial fast cache task cancellation-safe, and narrows hover recovery retries so they only continue when the pointer is actually near the menu bar.

## PR Type

- [x] Bugfix
- [ ] CI/CD related changes
- [ ] Documentation
- [ ] Feature
- [ ] Refactor
- [x] Performance improvement
- [ ] Test addition or update

- [ ] Yes
- [x] No

There were a few overlapping issues in the current behavior:

- stale hover or click tasks could continue running after cancellation and clear newer state
- duplicated `MenuBarItem` instances could lose the distinction between `nil sourcePID` and a resolved PID
- the initial fast cache retry loop could continue after cancellation
- after repeatedly toggling `Show on hover`, the first recovery path could become heavy and unreliable
- menu-open probing briefly cached both positive and negative results, which could reuse stale `false` results

Issue Number: N/A

## What is the new behavior?

This PR changes the behavior as follows:

- canceled show-on-click guard tasks now stop cleanly instead of disarming a newer guard
- hover tasks use identity tracking so older tasks cannot clear state belonging to newer ones
- duplicate `MenuBarItem` rebuilding now preserves `nil sourcePID`
- the initial fast cache task exits cleanly when canceled
- menu-open probing now only caches positive results
- re-enabling `Show on hover` triggers a short-lived recovery path, but retries are limited to cases where the pointer is actually near the menu bar

## PR Checklist

- [ ] I've run relevant tests and they pass

## Other information

This PR is intentionally focused on interaction/state reliability and startup-adjacent hover behavior.

I also noticed a separate startup-time performance issue where `Show on hover` and `Show on click` can still feel slow immediately after launch. That investigation is more involved, so I’m keeping it out of this PR and treating it as a separate follow-up.

## Notes for reviewers

The most important areas to review are:

- async task/state handling in `HIDEventManager`
- `sourcePID` preservation for duplicated `MenuBarItem` instances
- the narrowed hover rearm behavior after toggling `Show on hover`

stonerl:
@Skyearn I marked this as draft since PR #420 has some significant changes that will require modifications to this PR.

Skyearn:
> @Skyearn I marked this as draft since PR #420 has some significant changes that will require modifications to this PR.

I found it. I git pull found that there was a file with more than 6,000 lines of changes.

stonerl:
Wait for PR #420 to hit development, and then we can rebase on that. It completely drops macOS 14/15 support, so some of your changes won't apply to it, since we won't need any `#available(macOS 26.0, *)` and so on anymore.

stonerl:
@Skyearn you can now rework this PR. Not sure what is the best way for you. PR #420 has been merged into development and Thaw is now macOS 26 only which should make it easier for you since you can ignore all exceptions for macOS 14/15 since the code has been cleaned up and no code for older macOS verisons remained.

Skyearn:
> @Skyearn you can now rework this PR. Not sure what is the best way for you. PR #420 has been merged into development and Thaw is now macOS 26 only which should make it easier for you since you can ignore all exceptions for macOS 14/15 since the code has been cleaned up and no code for older macOS verisons remained.

👌

Skyearn:
@stonerl This is still the same set of changes as before.

After syncing with the latest upstream `development`, there was one conflict in `MenuBarItem.swift`. The conflict came from upstream moving to the new macOS 26-only single `getMenuBarItems(...)` path, while this PR previously added the `resolveSourcePID` fast-path control on top of the older experimental/legacy split.

I resolved it by keeping the upstream single-path implementation and reapplying the `resolveSourcePID` behavior there:

- `resolveSourcePID: false` still builds a lightweight item cache without resolving source PIDs
- `resolveSourcePID: true` keeps the full source PID resolution behavior
- unresolved `sourcePID` values are still preserved as `nil` instead of being implicitly replaced with `ownerPID`

So the intent of this PR is unchanged; the implementation was just adapted to the new macOS 26-only menu item path.

stonerl:
I'll do a review later. Thank you for you contributions 🙏

stonerl:
@coderabbitai resume

PR #423, opened 2026-04-21
perf: optimize startup performance for trigger-based components (showonhover/showonclick)

Skyearn:
## What does this PR do?

This PR optimizes the startup performance of the Thaw UI library by refining the initialization logic for components that use `showonhover` and `showonclick` triggers (such as Popover, Tooltip, and Dropdown). 

The core improvement involves **lazy-mounting the overlay content**, ensuring that hidden elements are not eagerly rendered into the DOM during the initial application bootstrap.

## PR Type

- [ ] Bugfix
- [ ] CI/CD related changes
- [ ] Documentation
- [ ] Feature
- [ ] Refactor
- [x] Performance improvement
- [ ] Test addition or update

- [ ] Yes
- [x] No

Previously, components with `showonhover` or `showonclick` behaviors would often render their hidden overlay content into the DOM immediately upon startup. For pages with a large number of these components (e.g., a long list of tooltips or menus), this resulted in:
1. Increased initial DOM depth and memory usage.
2. Slower WASM instantiation and hydration times in Leptos.
3. Unnecessary CPU cycles spent on calculating styles for hidden elements.

## What is the new behavior?

- **Lazy Rendering**: Overlay content is now only rendered and mounted to the DOM when the trigger event (hover or click) actually occurs for the first time.
- **Improved Hydration**: Reduced the number of reactive nodes that Leptos needs to track during the initial hydration phase.
- **Resource Efficiency**: Significantly lowered the initial memory footprint and improved the Time to Interactive (TTI) for heavy UI pages.

## PR Checklist

## Other information

N/A

### Notes for reviewers

Please verify if there are any side effects when these components are used inside a `Transition` or `Suspense` block, especially regarding the timing of the first-time hover/click interaction.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved hover delay handling to prevent redundant restarts.
  * Enhanced app focus detection on startup.

* **Performance**
  * Optimized menu-bar item hit testing with caching.
  * Added startup cache warming for faster initialization.
  * Improved menu-open detection with result caching and concurrency control.

### DIFF
--- a/Thaw/Events/HIDEventManager.swift
+++ b/Thaw/Events/HIDEventManager.swift
@@ -42,6 +42,10 @@ final class HIDEventManager: ObservableObject {
     /// The currently pending show-on-hover delay task.
     private var hoverTask: Task<Void, any Error>?
 
+    /// The currently pending hover action, used to avoid restarting the same
+    /// delay window on every small mouse move inside the same region.
+    private var pendingHoverAction: HoverAction?
+
     /// The pending task that clears the temporary show-on-click guard.
     private var clickTask: Task<Void, Never>?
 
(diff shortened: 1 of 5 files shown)


Skyearn:
wrong base repo select, merged into Skyearn:development.

Issue #430, opened 2026-04-22
[Bug]: High CPU usage


Andrewzhang00GH:
### Search for Similar Reports

- [x] I have searched existing issues for similar reports

### Description

<img width="1152" height="205" alt="Image" src="https://github.com/user-attachments/assets/4e863e75-1806-4af6-aedf-026dceda9f2b" />
Don't know how, but thaw became the highest CPU usage app (the usage is way too much).
It stays for a while but finally reduced to 3.1%.
Pls try to make it as light as possible.
This is a perfect app enhancing macOS, it is going to run at background like a system service for me, I need it to be efficient!

### Steps to Reproduce

Sorry but I don't know, I just saw the "stats app" that menu bar performance monitor app, shows CPU usage went up to 50+% and I'm doing almost nothing.
So I took a look at it and found Thaw is the top user.

### App Version

Version 1.2.0 (30)

### macOS Version

26.4.1

### Monitor Setup

Multi-monitor Setup

stonerl:
Would you kindly try the latest beta and see if the issue is fixed there?

diazdesandi:
Any update? @Andrewzhang00GH

stonerl:
Since I cannot reproduce this with the latest development build, I'm closing this one for now.

PR #482, opened 2026-04-29
fix(rehide): prevent appHasVisiblePopup spin on app switches

beantownbytes:
## What does this PR do?

Fixes a high-CPU regression where `rehideTemporarilyShownItems` could pin the main thread during rapid app-switching by repeatedly enumerating every on-screen window via SkyLight and never being able to clear stuck temp-shown contexts.

## PR Type

- [x] Bugfix
- [x] Performance improvement

- [x] No

Issue Number: #430

`MenuBarItemManager.TemporarilyShownItemContext.appHasVisiblePopup()` is the fallback path used when `shownInterfaceWindow` was never captured (or has gone away) and the 2 s grace period has expired. Two issues compound:

1. The popup test matched any window owned by the same PID whose layer was `> normalWindow` — including the temporarily-shown item itself at status-window level. So `isShowingInterface` returned `true` forever for any context that hit the fallback path, blocking rehide and re-arming the 3 s deferral timer indefinitely.
2. `runRehideTimer` subscribes to `NSWorkspace.shared.publisher(for: \.frontmostApplication)` with no debounce, so each Cmd-Tab queues another rehide call. Each call does a fresh system-wide `SLWindowListCreateDescriptionFromArray` (~1.5 s on my machine) on the main thread.

A 10 s `sample` taken during rapid app-switching with a stuck context showed `rehideTemporarilyShownItems → appHasVisiblePopup → SLWindowListCreateDescriptionFromArray` accounting for ~97 % of main-thread CPU.

## What is the new behavior?

- `appHasVisiblePopup` matches only `kCGPopUpMenuWindowLevel` (and the `-1` variant macOS sometimes uses). Status-level and main-menu-level windows are excluded — those are menu bar items themselves, not popups. The broad `> normalWindow` catch-all is gone; it picked up floating panels and modal levels that aren't popups either.
- The `frontmostApplication` publisher is debounced 200 ms so a burst of Cmd-Tabs collapses into one rehide attempt instead of queuing a Task per change.

## PR Checklist

- [ ] I've run relevant tests and they pass *(no automated tests exist for this code path)*
- [ ] I've verified localized strings work correctly (i18n/l10n changes) *(no localized strings touched)*

### Verification

- Repro: open Thaw Bar → click a hidden item (Amphetamine in my case) → click away → Cmd-Tab through 10–15 apps.
- Before: Thaw pinned the main thread for the duration of the cycle; the stuck context survived indefinitely.
- After: rehide completes within the grace window; CPU stays flat during Cmd-Tab spam.
- Sanity checks still pass: hidden items still show in the IceBar; clicking a hidden item still opens the popup and rehides correctly afterward; Mission Control transitions still work.

### Test environment

- macOS 26.4.1 (25E253), Xcode 16.4, multi-monitor.
- Built Release with `xcodebuild ... -configuration Release`, ad-hoc signed, daily-driven for the verification window.
- `swiftformat .` + `swiftlint lint` clean on the modified file.

### Notes for reviewers

The fast path of `isShowingInterface` (lines 2925–2942) still matches `statusWindow` and `mainMenuWindow` levels for the *tracked* `shownInterfaceWindow`. I left that alone — it operates on a single specific window the click captured rather than enumerating, so it's cheap and the existing semantics there have been stable. The fix is scoped to the slow fallback path that drives the storm.

If you'd prefer a more conservative fix (e.g. just exclude `statusWindow` from the original check while keeping the `> normalWindow` catch-all), I can split this. Tightening to `popUpMenuWindow` only matches what menu bar item popups actually are on macOS — but it's a behavior narrowing for any app that puts an item-triggered overlay at a non-menu level, and I haven't audited every menu bar app for that.

### DIFF
--- a/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
+++ b/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
@@ -2954,23 +2954,26 @@ extension MenuBarItemManager {
         }
 
         /// Checks whether the item's owning application has any visible
-        /// popup, menu, or overlay window on screen.
+        /// popup-menu window on screen.
+        ///
+        /// Only matches the pop-up menu level (the level macOS uses for
+        /// menus opened from menu bar items). Status-level and main-menu
+        /// level windows are excluded because those are the menu bar items
+        /// themselves — including the temporarily-shown item we're
+        /// tracking — not popups created by clicking them. A liberal
+        /// "above normal" match was previously used as a catch-all, but
+        /// it matched floating panels, modal levels, and other unrelated
+        /// app windows, keeping `isShowingInterface` true indefinitely
+        /// and preventing rehide.
         private func appHasVisiblePopup() -> Bool {
             let windows = WindowInfo.createWindows(option: .onScreen)
+            let popUpLevel = CGWindowLevelForKey(.popUpMenuWindow)
             return windows.contains { window in
                 guard window.ownerPID == sourcePID else {
                     return false
                 }
-                // Menu-level or status-level windows are popups.
-                if window.isMenuRelated {
-                    return true
-                }
-                // Above-normal layer windows (overlays, popovers) that
-                // belong to the app also count.
-                if window.layer > CGWindowLevelForKey(.normalWindow) {
-                    return true
-                }
-                return false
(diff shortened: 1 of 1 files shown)


stonerl:
LGTM

PR #490, opened 2026-04-29
fix: stabilize hidden menu items during capture

Yukaii:
## Summary
- classify transient macOS capture indicators and avoid moving or capturing over them
- tighten hidden Ice Bar click targeting so hidden rows are moved out instead of clicking same-looking visible items
- improve temporary menu popup detection and cursor restoration around synthetic menu bar interactions

## Root Cause
macOS capture indicators and Control Center-hosted menu extras can appear with transient or generic menu bar window metadata. Thaw could move/click/capture while those windows overlapped the hidden-section controls, leaving stale cached backgrounds or clicking the wrong visible menu item.

## Validation
- xcodebuild test -project Thaw.xcodeproj -scheme Thaw -configuration Debug -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO -only-testing:ThawTests/MenuBarItemTagTests\n

### DIFF
--- a/Thaw/MenuBar/IceBar/IceBar.swift
+++ b/Thaw/MenuBar/IceBar/IceBar.swift
@@ -208,7 +208,7 @@ final class IceBarPanel: NSPanel {
         cacheTask?.cancel()
         cacheTask = Task { [weak appState] in
             guard let appState else { return }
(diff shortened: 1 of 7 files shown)


stonerl:
@Yukaii thank you for opening this PR. let me know when it is ready to merge.

PR #522, opened 2026-05-04
fix(itemManager): restore profile layout after settling and on app relaunch

nightah:
## What does this PR do?

Fixes two related bugs that left menu bar icons in the wrong section or order: (1) on initial launch of Thaw the saved profile layout was not applied because `applyProfileLayout` raced the startup settling period, and (2) when an app whose icon belongs to the active profile was quit and relaunched, its icon was not moved back to the saved profile position.

## PR Type

- [x] Bugfix
- [ ] CI/CD related changes
- [ ] Documentation
- [ ] Feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] Test addition or update

- [ ] Yes
- [x] No

N/A

Issue Number: N/A

On initial launch, `applyProfileLayout` was dispatched immediately by `ProfileManager.performSetup` and ran while `isInStartupSettling` was `true`. Concurrent `cacheItemsRegardless` cycles took the settling guard branch, skipped restore, and absorbed every current item into `profileSortedItemIdentifiers`. The LCS move loop's effects were silently shadowed and the post-settling late-arrival re-sort path then filtered those items out as already sorted, so the saved profile layout was never actually applied for items present at startup. Separately, because `uniqueIdentifier` is `namespace:title` (windowID-independent so persisted custom names survive restarts), a relaunched app reused the same identifier and was filtered out of `newProfileItems` by the subtraction against `profileSortedItemIdentifiers`, so `scheduleProfileResort` never fired even though the `NSStatusItem` had been freshly registered at a position that did not match the saved profile.

## What is the new behavior?

`applyProfileLayout` now awaits `startupSettlingTask.value` before touching any state, so it runs once against the settled cache produced by the final `cacheItemsRegardless` passes; `activeProfileLayout` stays `nil` for the duration of settling, making the absorb branch a no-op. `cacheItemsRegardless` now also detects relaunched profile items by scanning for items whose `windowID` is absent from `previousWindowIDs` and whose `uniqueIdentifier` is in `activeProfileItemIdentifiers`, and drops those identifiers from `profileSortedItemIdentifiers` so the existing late-arrival check picks them up and triggers `scheduleProfileResort`. Manual drags via Layout Bar continue to persist because their `windowID` does not change.

## PR Checklist

## Other information

Diagnostic log lines to watch when reproducing: `applyProfileLayout: waiting for startup settling to end` on cold start (only when settling is still active at apply time), and `Profile re-sort: detected N relaunched profile item(s) with fresh windowID: [...]` followed by `Profile re-sort: re-applying layout for late-arriving items` when a profiled app is quit and relaunched.

### Notes for reviewers

Two focused changes in `Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift`: a six-line settling gate at the top of `applyProfileLayout`, and a pre-check inside `cacheItemsRegardless` that prunes stale entries from `profileSortedItemIdentifiers` for items whose `windowID` just appeared. The absorb branch at the existing settling guard is now a no-op under normal flow but is intentionally left in place as a defensive guard against future call paths that might set `activeProfileLayout` outside `applyProfileLayout`. No changes to `ProfileManager` were required because gating inside `applyProfileLayout` covers all entry points (initial launch, display switch, focus filter, manual profile switch).

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate re-sorting of menu bar items after app relaunch by distinguishing relaunch-induced changes from genuinely new items.
  * Improved reliability of profile layout application by blocking profile changes until startup settling finishes, avoiding intermittent or premature layout restores.

### DIFF
--- a/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
+++ b/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
@@ -1657,6 +1657,35 @@ extension MenuBarItemManager {
         if activeProfileLayout != nil,
            !activeProfileItemIdentifiers.isEmpty
         {
+            // App-relaunch detection: uniqueIdentifier is namespace:title
+            // (windowID-independent and stable across restarts), so a
+            // relaunched app keeps the same identifier and would be filtered
+            // out of newProfileItems by profileSortedItemIdentifiers below.
+            // A fresh windowID for a profile-tracked item means the app
+            // re-registered its NSStatusItem at whatever position macOS
+            // chose, which is usually not the saved profile position. Drop
+            // such identifiers from the sorted snapshot so the existing
+            // late-arrival path picks them up.
+            let previousWindowIDSet = Set(previousWindowIDs)
+            if !previousWindowIDSet.isEmpty {
+                let relaunchedIdentifiers = Set(
+                    items
+                        .filter { item in
+                            !item.isControlItem
+                                && !previousWindowIDSet.contains(item.windowID)
+                                && activeProfileItemIdentifiers.contains(item.uniqueIdentifier)
+                        }
+                        .map(\.uniqueIdentifier)
+                )
+                if !relaunchedIdentifiers.isEmpty {
+                    let staleSorted = relaunchedIdentifiers.intersection(profileSortedItemIdentifiers)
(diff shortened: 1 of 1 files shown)


stonerl:
LGTM

PR #552, opened 2026-05-08
fix: keep menu bar spacing global

### LINKED ISSUE #551: Menu bar item spacing is modeled as per-display despite being global
## Summary
Menu bar item spacing is currently stored on `DisplayIceBarConfiguration` and exposed in the Displays pane as if it were a per-display setting.

## Problem
macOS reads `NSStatusItemSpacing` and `NSStatusItemSelectionPadding` from the by-host global domain. The current implementation stores a different offset per display, then rewrites the single host-wide value when the active display/configuration changes. That makes the UI misleading and can trigger unnecessary menu bar app relaunch waves from display configuration changes.

## Expected behavior
Menu bar item spacing should be modeled as a global General setting, matching the underlying macOS preference scope. Display settings should only contain settings that are actually display-scoped.

## Proposed fix
Move `itemSpacingOffset` back to General settings/profile snapshots, remove it from `DisplayIceBarConfiguration`, and keep the explicit Apply flow that relaunches affected menu bar apps only when the user applies the global spacing value.

CamilleGuillory:
## Summary
- Moves menu bar item spacing back to General settings, matching macOS' host-wide `NSStatusItemSpacing` / `NSStatusItemSelectionPadding` behavior.
- Removes spacing from per-display configuration/UI and stops display configuration changes from triggering spacing relaunch waves.
- Keeps profile snapshot support for global spacing and updates tests for the General-settings model.

Fixes #551

## Test plan
- `ReadLints` on touched Swift files
- `git diff --check`
- Not run: `swiftformat` (not installed in this environment)
- Not run: `xcodebuild test -project Thaw.xcodeproj -scheme Thaw -derivedDataPath Build/ -destination 'platform=macOS' -enableCodeCoverage YES CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO` because the active developer directory is Command Line Tools, not full Xcode

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Global "Item Spacing Offset" added to General Settings with a new "Spacing Options" slider, plus Apply and Reset actions.
  * Resetting settings now restores the global spacing offset to its default.

* **Removed Features**
  * Per-display item spacing controls removed from Display Settings.

* **Tests**
  * Updated and added tests to cover the new global spacing field, decoding/clamping, and legacy compatibility.

### DIFF
--- a/Thaw/MenuBar/IceBar/IceBar.swift
+++ b/Thaw/MenuBar/IceBar/IceBar.swift
@@ -360,7 +360,7 @@ private struct IceBarContentView: View {
     }
 
     private var itemSpacing: CGFloat {
-        let offset = displaySettings.configuration(for: screen.displayID).itemSpacingOffset
+        let offset = appState.settings.general.itemSpacingOffset
         return max(0, CGFloat(offset).rounded())
     }
 
--- a/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
+++ b/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
@@ -924,9 +924,9 @@ final class MenuBarItemManager: ObservableObject {
 
         // Boot race: a cold (performSetup) or expected-set settling must
         // not be torn down by a transient preflight that the boot path
-        // also kicks off (DisplaySettingsManager.applyActiveDisplaySpacing,
-        // ProfileManager.layoutTask). Preserve the merged expected set so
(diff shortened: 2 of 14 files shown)


stonerl:
Closing this one for now, since it is not what we want as discussed in #551

PR #569, opened 2026-05-12
fix(display): skip relaunch wave on resolution-only screen-parameters changes

### LINKED ISSUE #551: Menu bar item spacing is modeled as per-display despite being global

nightah:
## What does this PR do?

Stops `DisplaySettingsManager.applyActiveDisplaySpacing` from firing on `NSApplication.didChangeScreenParametersNotification` events that only reflect resolution or other-parameter changes on the same display, so a resolution flip no longer risks triggering a full menu-bar-app relaunch wave.

## PR Type

- [x] Bugfix
- [ ] CI/CD related changes
- [ ] Documentation
- [ ] Feature
- [ ] Refactor
- [ ] Performance improvement
- [x] Test addition or update

- [ ] Yes
- [x] No

N/A

PR #529 (commit `1ed2ce67`) wired the spacing apply to `didChangeScreenParametersNotification`. That notification fires on every screen-parameter event, including resolution changes, lid open/close, and GPU/sleep transitions on an already-connected display. Each fire armed a preflight settling window and ran `applyOffset`. When the on-disk `NSStatusItem` byHost values disagreed with the target (Control Center restart, sleep/wake clobber, mid-flight oscillation across profiles) the call escalated to a full relaunch wave even though the user had only changed a setting of the display, not changed to a different one.

Issue Number: #551

## What is the new behavior?

`DisplaySettingsManager` now caches the active menu bar display UUID at every `applyActiveDisplaySpacing` call and the `didChangeScreenParametersNotification` sink consults a pure static predicate `shouldSkipSpacingApply(currentActiveDisplayUUID:lastAppliedActiveDisplayUUID:)` before falling through. When the UUIDs match (resolution change, lid event, or GPU transition on the same display) the sink logs "Active menu bar display unchanged" and returns. When they differ (laptop docked, active menu bar moved to another monitor, first event after launch) the existing apply path runs as before. The `$configurations` sink and external Settings-URI path are unchanged because those reflect deliberate user intent and already rely on the no-op guard inside `MenuBarItemSpacingManager.applyOffset` to short-circuit when target equals on-disk.

## PR Checklist

## Other information

Adds `ThawTests/DisplaySettingsManagerSpacingGateTests` with eight cases pinning the predicate (matching, differing, either-nil, both-nil, repeated-stable) and the field semantics on a fresh `DisplaySettingsManager` instance. The file is picked up automatically by the `ThawTests` synchronized-folder reference, so `Thaw.xcodeproj/project.pbxproj` requires no edits.

### Notes for reviewers

The predicate is intentionally a static pure function so it can be unit-tested without `AppState`, `Bridging`, or live `NSScreen` state. End-to-end driving of the notification sink would depend on the host machine's display topology and the 1 s `.debounce`, so the manual verification steps below cover that surface instead:

1. Change the active display's resolution in System Settings: log shows "Active menu bar display unchanged"; no relaunch wave.
2. Move the active menu bar to a different physical display: apply runs as before.
3. Drag the per-display spacing slider in Settings > Displays: configurations sink fires and re-applies.
4. Drive a per-display setting via the Settings URI scheme: external path runs as before.
5. Unplug/replug the active display: `Bridging.getActiveMenuBarDisplayUUID()` returns the same UUID, so no redundant wave on reconnect.

Fixes #551

## Release Notes

* **Bug Fixes**
  * Improved display spacing handling to prevent unnecessary recalculations when screen parameters change.

* **Tests**
  * Added comprehensive test coverage for display spacing behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/stonerl/Thaw/pull/569)

### DIFF
--- a/Thaw/Settings/Models/DisplaySettingsManager.swift
+++ b/Thaw/Settings/Models/DisplaySettingsManager.swift
@@ -40,6 +40,12 @@ final class DisplaySettingsManager: ObservableObject {
     /// active-display configuration changes. Held weakly to avoid retain cycles.
     private weak var appState: AppState?
 
+    /// UUID of the active menu bar display the last time spacing was applied.
+    /// Used to skip didChangeScreenParametersNotification fires that only
+    /// reflect a resolution or other-parameter change on the same display.
+    /// Internal access so unit tests in ThawTests can seed and assert it.
+    var lastAppliedActiveDisplayUUID: String?
+
     /// Performs the initial setup of the manager.
     func performSetup(with appState: AppState) {
         self.appState = appState
@@ -162,6 +168,14 @@ final class DisplaySettingsManager: ObservableObject {
                 guard let self else { return }
                 diagLog.info("Screen parameters changed — \(NSScreen.screens.count) screen(s) connected")
                 captureCurrentlyConnectedDisplays()
+                let currentUUID = Bridging.getActiveMenuBarDisplayUUID()
+                if Self.shouldSkipSpacingApply(
+                    currentActiveDisplayUUID: currentUUID,
+                    lastAppliedActiveDisplayUUID: lastAppliedActiveDisplayUUID
+                ) {
+                    diagLog.info("Active menu bar display unchanged (\(currentUUID ?? "nil")); skipping spacing apply")
+                    return
+                }
                 applyActiveDisplaySpacing(reason: "screenParametersChanged")
             }
             .store(in: &c)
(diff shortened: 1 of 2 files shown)


stonerl (approved):
LGTM

PR #577, opened 2026-05-13
feat: add global and per-profile pre/post apply script hooks

### LINKED ISSUE #567: [Feature Request]: Ability to call bash scripts as hook on profile switch

### Description

The new profile feature is awesome, especially notch detection and profile switching on different monitor combinations.

There are other menu bar apps that do not do this kind of environment detection, but do provide APIs/methods for switching their config profiles.

Would be great for the profile switch event to also drive the execution of a user written shell/bash script, to manage the profile/settings switching when thaw executes a profile change.

For background info, my use case is switching istat menus profiles on notch/non-notch displays. Where non notch will show significantly more system info, while notch displays might only show minimalistic bars/dials.
Istat only supports profile changes the following ways:
- plist editing
- menu driven import/export,
- file switching

So programatic options are really just, bash/applescript.

### Additional Information

_No response_

nightah:
## What does this PR do?

Adds pre/post profile-apply hooks that run user-supplied shell scripts (any executable with a shebang) or AppleScript files (`.scpt`, `.applescript`, `.scptd`) around every Thaw profile switch, in two scopes: Global (every apply) and Per-Profile (only when its owning profile is applied). All apply paths funnel through `ProfileManager.applyProfile`, so the hook stages cover the manual Apply button, the profile hotkey, the menu bar and control item context menus, the display auto-switch, and macOS Focus Filter activation.

## PR Type

- [ ] Bugfix
- [ ] CI/CD related changes
- [ ] Documentation
- [x] Feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] Test addition or update

- [ ] Yes
- [x] No

There is no way to run a script when Thaw switches profiles. Users with sister menu bar apps that don't expose a profile-switch API (the issue cites iStat Menus, which responds to `defaults write` and AppleScript but has no native sync mechanism) have no programmatic hook to keep those apps' configurations in step with Thaw's display-driven or focus-filter-driven transitions.

Issue Number: #567

## What is the new behavior?

A new Hooks section appears in the Automation settings pane, with two subsections. Global Hooks holds pre and post rows that fire on every apply. Per-Profile Hooks holds a profile picker plus pre and post rows for the selected profile; switching the picker swaps the rows to that profile's hooks. Each hook row uses `NSOpenPanel` to pick a script file, a stepper to set a wall-clock timeout (1-300 seconds, default 5), and an enable toggle. Hooks can be paused without losing the picked path by toggling enable off. Validation warnings surface in the row when the chosen file is missing or, for non-AppleScript files, lacks the executable bit.

Hooks run in symmetric nested order around the apply: `Global pre -> Profile pre -> [Thaw applies the snapshot, spacing offset, and layout] -> Profile post -> Global post`. With only one scope configured, the missing stages collapse out (`Global pre -> [apply] -> Global post`, or `Profile pre -> [apply] -> Profile post`). Profile hooks belong to the incoming profile, not the outgoing one; teardown for the outgoing profile is best expressed as a Global post that branches on `THAW_PREVIOUS_PROFILE_NAME`. Hooks are blocking: the layout task awaits each stage before proceeding, capped at the configured timeout. Failures and timeouts log to `DiagLog` under category `HookRunner` but never abort the apply, so a broken script can't lock the user out of switching profiles. Cancelled applies (rapid back-to-back profile switches) propagate `Task.isCancelled` through the hook stages.

Every hook process inherits Thaw's environment merged with `THAW_HOOK_PHASE` (`Pre` or `Post`), `THAW_HOOK_SCOPE` (`Global` or `Profile`), `THAW_PROFILE_ID` (UUID of the incoming profile), `THAW_PROFILE_NAME` (user-visible name), `THAW_PREVIOUS_PROFILE_ID` (UUID of the previously active profile, empty if none), and `THAW_PREVIOUS_PROFILE_NAME` (name or empty). For example, a one-line bash pre-hook that mirrors the switch into iStat Menus: `defaults write com.bjango.istatmenus5 ActiveProfile -string "$THAW_PROFILE_NAME"`. An AppleScript post-hook reads the same values via `system attribute "THAW_PROFILE_NAME"` (osascript does not expose Unix env vars through the standard AppleScript environment accessor). A shell hook can branch on phase with `case "$THAW_HOOK_PHASE" in Pre) ...; Post) ...; esac`.

Implementation: `HookScript` and `ProfileAutomation` are new in `Thaw/Settings/Models/HookScript.swift`. Global hooks persist via two new `Defaults` keys (`globalPreProfileHook` / `globalPostProfileHook`) as JSON-encoded `Data`. Per-profile hooks live inside the profile's JSON via a new optional `automation: ProfileAutomation?` field on `Profile` and `ProfileContent`, decoded with `decodeIfPresent` so existing profile files on disk are unaffected; profile export and import already serialise the whole `Profile`, so per-profile hooks round-trip through `exportProfile` and `importProfile` automatically. `HookRunner` in `Thaw/Utilities/HookRunner.swift` wraps `Process()` with `Pipe`s and races termination against a `Task.sleep` timeout via `withThrowingTaskGroup`; AppleScript files route through `/usr/bin/osascript <path>`, other paths launch directly and are checked for the executable bit at run time. `ProfileManager.applyProfile`'s previously-synchronous snapshot push (general / advanced / hotkeys / display config / appearance / item names / new-items placement) now runs inside the `layoutTask` after the pre-hook stages so a pre-hook can complete before Thaw mutates state, and `applyProfile` gained a `previousProfileID` parameter that all seven call sites pass by reading `activeProfileID` before they overwrite it.

## PR Checklist

## Other information

Manually verified the following scenarios on a multi-display setup, against the local debug build with diagnostic logging enabled:

- Profile post-hook on Profile A writes its log only after `await profileManager.layoutTask?.value` returns.
- All four hooks (Global pre/post + Profile pre/post) appending to one file produce the lines `globalPre, profilePre, profilePost, globalPost` in that order.
- `sleep 60` hook with timeout `5` logs `HookError.timedOut`; Apply button re-enables.
- `.sh` without `+x` logs `HookError.notExecutable`; apply still completes.
- `.scpt` hook routes through `osascript` (visible in Activity Monitor); no `+x` required on the script.
- `env | grep THAW_` confirms all six env vars are set, with `THAW_PROFILE_NAME` and `THAW_PREVIOUS_PROFILE_NAME` reflecting the transition.
- Export -> delete -> import round-trip preserves per-profile hooks.
- No-hook applies short-circuit on `hook == nil || !hook.isEnabled`; latency matches the pre-PR baseline.
- Back-to-back applies: the first layout task cancels and its post-hooks do not run.

### Notes for reviewers

- The hook execution model is "log and continue" by design (per the issue discussion). Script failures or timeouts never abort an apply, so a broken hook can't lock the user out of switching profiles. The failure path is centralised in `HookRunner.runIfEnabled`, so adding an `abortOnFailure` flag later is a small change.
- `previousProfileID` is threaded through `applyProfile` rather than read from `activeProfileID` inside the function, because every call site overwrites `activeProfileID` before invoking `applyProfile`. Reading inside the function would always observe the new value.
- `Process()` inherits Thaw's full environment by design and merges the `THAW_*` vars on top, so scripts see the user's `PATH`, locale, and so on without extra configuration.
- The Per-Profile picker lives in the Automation pane rather than inside each profile row in the Profiles pane. The Profiles pane stays focused on layout / display association / apply controls; configuring scripted side-effects fits next to the Settings URI Scheme controls already in Automation.

Closes #567

## Summary by CodeRabbit

* **New Features**
  * Global and per-profile automation hooks (pre/post) with UI for selecting scripts, enable/disable, and timeout.
  * Hooks execution with environment variables exposing profile and phase information.
  * Localized UI strings and settings for hook management.

* **Bug Fixes / Behavior**
  * Profile apply flow now preserves and provides the previous profile context to hooks so scripts receive prior-profile info.

* **Validation**
  * File existence and executability warnings for selected scripts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/stonerl/Thaw/pull/577)


nightah:
Here's a short video illustrating the change. This shows how you can have both global and profile hooks set.


https://github.com/user-attachments/assets/8f7970ca-4238-49d8-bca3-698544dff792

Issue #481, opened 2026-04-28
[Bug]: full screen app with hidden menu misclick


kn666:
### Search for Similar Reports

- [x] I have searched existing issues for similar reports

### Description

If another app is fullscreen, menubar is hidden, clicks at the menu area level produce unwanted reactions. 

### Steps to Reproduce

1. Open full-screen app with the menu-bar hidden
as is: click at the top row - thaw reacts
to be: clicks should be ignored

### App Version

1.2.0 30

### macOS Version

26.4.1

### Monitor Setup

Single Monitor

plasticrake:
I had this issue but switching to the beta seems to have fixed it.

kn666:
Thanks. For me it works half-way:
1. clicking on icons area - ok
2. clicking on empty space area - not ok, shows hidden icons

stonerl:
@kn666 is this on a display w/o a notch? Or do you have the menubar below the notch?

kn666:
@stonerl yes, this is w/o a notch in fullscreen, with menu hidden.

stonerl:
Thanks. That helps pinpoint the problem.

nightah:
Could you please give clear reproduction steps?

If I have my Menu Bar settings set to `Automatically hide and show the menu bar: In Fullscreen` when I hover to the top of the fullscreen window, the menu bar drops down over the fullscreen application.

Firstly, I would recommend attempting to reproduce this with the latest 2.0.0-beta7. If you can reproduce it, please specify the application that you're full-screening, the relevant menu, MacOS menu bar settings you have set, and whether you're left or right clicking on the empty menu bar space.

Also, having an image of your settings for the `Show on ...` settings from Thaw would be useful too:

<img width="3096" height="1215" alt="Image" src="https://github.com/user-attachments/assets/70599ae5-d23a-4a7d-9b6a-7ee9872b3248" />

kn666:
https://github.com/user-attachments/assets/0fa8b37e-6098-480c-9962-6d11e58bcc15

<img width="791" height="529" alt="Image" src="https://github.com/user-attachments/assets/401c8140-fbec-4aa2-94dc-6e09ad7b7155" />

Checked in Version 2.0.0-beta.8 (39). Settings screen attached. 

Finder (or any app) opened full screen on external monitor (no notch) connected in clam-shell mode. Clicking in hidden menu area shows hidden icons. See the attached video

diazdesandi:
[...]

PR #584, opened 2026-05-14
fix(menubar): defer reveal and right-click paths to third-party widgets

nightah:
## What does this PR do?

Extends Thaw's menu bar event handling so all four reveal paths (`handleShowOnClick`, `handleShowOnHover`, `handleShowOnScroll`, `handleSecondaryContextMenu`) defer to third-party menu bar widgets under the cursor, using a new system-wide accessibility hit-test. Also makes the right-click secondary context menu cancellable so a dismiss click inside its 100ms reveal delay can no longer race the reveal, and adds a foreign-pop-up probe so widgets that open their own context menu at `kCGPopUpMenuWindowLevel` no longer have it dismissed by Thaw's secondary menu.

## PR Type

- [x] Bugfix
- [ ] CI/CD related changes
- [ ] Documentation
- [ ] Feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] Test addition or update

- [ ] Yes
- [x] No

N/A

Clicks, hover, scroll, and right-click in the menu bar y-band that visually land on a third-party widget (notch overlay applications such as those that sit above the menu bar) trigger Thaw's section reveal or secondary context menu, because the existing geometric `isMouseInsideEmptyMenuBarSpace` check treats any non-status-item region as empty space. The widget's own pop-up menu, when one opens, is dismissed by Thaw's secondary context menu after the 100ms reveal delay. A rapid dismiss click within that 100ms window can also re-pop the secondary context menu after the user has already moved on.

Issue Number: Refs #481

## What is the new behavior?

A new system-wide AX hit-test `isCursorOverForeignWidgetUIElement` resolves the actual UI element under the cursor and defers when the element is owned by a non-Thaw, non-WindowServer, non-SystemUIServer process whose role is not `menuBar`, `menu`, `menuItem`, or `menuBarItem`. All four reveal paths consult it after the cheap geometric `isMouseInsideEmptyMenuBarSpace` check passes, so the AX cost only kicks in for clicks that would otherwise have been treated as empty menu bar space. `handleSecondaryContextMenu` additionally probes `isForeignPopUpMenuOpen` for any non-Thaw window at `kCGPopUpMenuWindowLevel`, and now runs as a cancellable `Task` stored in `pendingSecondaryContextMenuTask`. Every mouse-down (left or right) cancels the pending Task at the top of `mouseDownMonitor` and the Task checks `Task.checkCancellation()` after its 100ms sleep. `AXHelpers.pid(for:)` is added as a small wrapper around `AXUIElementGetPid` to support the hit-test.

## PR Checklist

## Other information

Manually verified the following scenarios on an external display (which is where the regressions surface; the built-in display's reveal animation hides the y-band gap):

- Right-click on Thaw's chevron in fullscreen revealed mode: Thaw context menu fires.
- Right-click on phantom menu bar area in fullscreen (menu bar auto-hidden): event passes through to the underlying app's context menu, Thaw stays quiet.
- Right-click on truly empty Thaw bar space in fullscreen revealed mode: Thaw secondary context menu fires.
- Right-click on a notch overlay widget with its own pop-up menu: widget's menu opens and stays; Thaw's `isForeignPopUpMenuOpen` probe suppresses.
- Right-click on a notch overlay widget that does not open a pop-up menu: Thaw's AX hit-test suppresses, no Thaw menu fires.
- Rapid right-click followed by an immediate left-click to dismiss: cancellation prevents the queued reveal from re-popping.
- Left-click, hover, and scroll variants of the widget-deferral case all suppress via the shared `isCursorOverForeignWidgetUIElement` gate.

### Notes for reviewers

The widget-detection model is intentionally AX-based rather than window-bounds-based. Notch overlay applications create wide opaque-but-non-interactive window regions for animation purposes, so bounds-based heuristics either misclassify large empty stretches as the widget (when the threshold is loose) or miss the widget entirely (when it is tight). The AX hit-test asks the system what UI element is actually at the cursor, which gives the right answer without per-widget allowlists. AX queries run on the existing `AXHelpers.queue`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed context menu appearing over third‑party menu‑bar widgets.
  * Improved secondary (right‑click) context menu reveal timing with robust cancellation to avoid stale triggers.
  * Suppressed menu reveal when other pop‑up menus are open or when no menu‑bar items are visible.
  * Prevented show‑on‑hover and show‑on‑scroll from triggering when the cursor is over foreign widget UI elements, reducing unintended reveals.

[Review Change Stack](https://app.coderabbit.ai/change-stack/stonerl/Thaw/pull/584)

### DIFF
--- a/Shared/Utilities/AXHelpers.swift
+++ b/Shared/Utilities/AXHelpers.swift
@@ -48,4 +48,12 @@ enum AXHelpers {
     static func role(for element: UIElement) -> Role? {
         queue.sync { try? element.role() }
     }
+
+    static func pid(for element: UIElement) -> pid_t? {
+        queue.sync {
+            var pid: pid_t = 0
+            let result = AXUIElementGetPid(element.element, &pid)
+            return result == .success ? pid : nil
+        }
+    }
 }
--- a/Thaw/Events/HIDEventManager.swift
+++ b/Thaw/Events/HIDEventManager.swift
@@ -130,6 +130,11 @@ final class HIDEventManager: ObservableObject {
     /// The pending tooltip show task.
     private var tooltipTask: Task<Void, any Error>?
 
+    /// The pending secondary-context-menu reveal task. Holding a reference lets
+    /// a subsequent click (right or left) cancel the pre-100ms-delay reveal so
+    /// the menu doesn't pop up after the user has already dismissed or moved on.
+    private var pendingSecondaryContextMenuTask: Task<Void, any Error>?
+
     /// A Boolean value that indicates whether the manager is enabled.
     private var isEnabled = false {
         didSet {
@@ -180,6 +185,10 @@ final class HIDEventManager: ObservableObject {
         } else {
             return event
         }
+        // Cancel any pending secondary-context-menu task on every mouse-down so
+        // a dismiss click or a follow-up click can't be raced by a previously
+        // scheduled reveal that hasn't yet cleared its 100ms delay.
+        pendingSecondaryContextMenuTask?.cancel()
         switch event.type {
         case .leftMouseDown:
             // Check app menu first - if click is on app menu area, don't trigger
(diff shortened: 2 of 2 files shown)


diazdesandi:
.build

diazdesandi (approved):

PR #613, opened 2026-05-24
perf(capture): route window capture through ScreenCaptureKit to cut SkyLight leaks

nightah:
## What does this PR do?

Migrates Thaw's window capture path from the leaking SkyLight private API (`SLWindowListCreateImageFromArray`) to ScreenCaptureKit (`SCScreenshotManager.captureImage(contentFilter:configuration:)`) for every call site whose target windows fit within display bounds, and batches the residual SkyLight calls that genuinely cannot move to SCK. Eliminates ~79% of the `NSMutableDictionary` leaks reported by `leaks` on macOS 26.4.1 and reduces total leaked bytes by ~76%.

## PR Type

- [ ] Bugfix
- [ ] CI/CD related changes
- [ ] Documentation
- [ ] Feature
- [ ] Refactor
- [x] Performance improvement
- [ ] Test addition or update

- [ ] Yes
- [x] No

N/A. The public surfaces of all migrated functions are preserved. `refreshImages` only gains a new `viaSCK: Bool = false` parameter with a default that matches prior behaviour. `compositeCapture` and `individualCapture` become `async`, but they are `private` to `MenuBarItemImageCache`. The three color-sampler functions (`MenuBarManager.updateAverageColorInfo`, `IceBarColorManager.updateWindowImage`, `MenuBarSearchModel.updateAverageColorInfo`) keep their synchronous signatures and externally observable behaviour.

Every call to `Bridging.captureWindowsImage` (the wrapper around the private `SLWindowListCreateImageFromArray`) leaks one `CFMutableDictionary` inside `SLSWindowListCreateImageFromArrayProxying` via `CGRectCreateDictionaryRepresentation`. `leaks` on a fresh `MallocStackLogging=1` Thaw process attributed ~250 `NSMutableDictionary` instances (~42 KB of ~44 KB total leaked bytes) to that one allocation site, exercised by six distinct Thaw call sites: `MenuBarItemImageCache.individualCapture`, `compositeCapture`, and `refreshImages`, plus `MenuBarManager.updateAverageColorInfo`, `IceBarColorManager.updateWindowImage`, and `MenuBarSearchModel.updateAverageColorInfo`. The leak is inside Apple's SkyLight framework, so the dictionary cannot be released from Swift; the only levers are to stop calling SkyLight or to call it less.

Issue Number: N/A

## What is the new behavior?

Adds `Bridging.captureWindowsImageSCK` as an async, leak-free SCK replacement (using `SCContentFilter(display:including:)` + `SCScreenshotManager.captureImage(contentFilter:configuration:)`) and exposes it via `ScreenCapture.captureWindowsAsync` / `captureWindowAsync`. Migrates `MenuBarItemImageCache.compositeCapture` and `individualCapture` to the SCK path (both become `async`). Migrates the three color-sampler functions while preserving their sync signatures by wrapping the capture in a `Task`; the SCK call runs off the main actor and state mutation resumes on the owning `@MainActor` class. For `refreshImages`, where items live past the display's left edge and both SCK filter shapes fail (`-3812` for display-bounded, `-3811` for desktopIndependentWindow), `refreshImages` gains an opt-in `viaSCK: Bool = false` parameter and `runLiveRefreshLoop` is restructured: tick-global guards (`lastMoveOperationOccurred`, `isResettingLayout`) hoist out of the per-section loop, the `.visible` section refreshes individually via the leak-free SCK path, and `.hidden` plus `.alwaysHidden` items batch into a single SkyLight call per tick instead of two separate ones. Net result on testing across normal exercise: 122 leaks / 10.8 KB total versus the 507 leaks / 44 KB baseline; the remaining 53 `NSMutableDictionary` instances are the irreducible Apple-framework floor.

## PR Checklist

## Other information

Verified empirically rather than via unit tests, since the bug surfaces in `leaks` output not in functional assertions. Workflow: build with `xcodebuild -scheme Thaw-configuration Debug build`, launch with `MallocStackLogging=1`, exercise all four scenarios (menu bar idle, ThawBar visible-section, ThawBar hidden + always-hidden expanded, `Settings → Menu Bar Layout` open), then `leaks <pid> --fullStacks --groupByType` and inspect attribution. The diagnostic delta is the most useful artefact: before the change, dictionary leaks scale with total capture-call count across all six call sites; after, they scale only with the batched `refreshImages` SkyLight call frequency. Comments at `Thaw/Utilities/ScreenCapture.swift` and `Shared/Bridging/Bridging.swift` document the SCK display-bounds limitation and the specific error codes (`-3812` / `-3811`) so future maintainers don't re-investigate.

### Notes for reviewers

The `refreshImages` `viaSCK` parameter defaults to `false` deliberately so any future external caller keeps the SkyLight semantics (offscreen-capable). Only the in-loop call site in `runLiveRefreshLoop` opts in to `true`, and only for `.visible`. Per-section routing is by section enum rather than runtime bounds-check because the section identity is already known at the call site and items in `.hidden` / `.alwaysHidden` are off-screen by definition (control items observed at x = -3985 and x = -9245 respectively). Color-sampler `Task` wraps trade a small change in update timing for leak elimination: paths that previously ran `updateWindowImage(); updateColorInfo();` back-to-back will now read the *previous* cycle's `windowImage` until the next async capture lands. On the 5 s periodic refresh this is imperceptible, but it's worth a quick visual smoke when exercising IceBar adaptive backgrounds. Apple Feedback drafts for the SkyLight `CGRectCreateDictionaryRepresentation` leak, the AppKit `_regionForOpaqueDescendants` `CGRegion` leak, and a separate `AppIntents` `NSProgress` retain cycle (observed when System Settings materializes `SetFocusFilterIntent` metadata) are prepared and will be filed via Feedback Assistant; FB numbers will be added as in-code comments once issued.

## Summary by CodeRabbit

* **New Features**
  * Added an async ScreenCaptureKit-backed capture path for display-bounded window screenshots and per-display captures.

* **Bug Fixes**
  * Prevented stale async captures from overwriting current color state via generation-based invalidation.
  * Improved capture failure handling and reduced missed updates during display/config changes.

* **Refactor**
  * Migrated menu-bar and image-cache refresh flows to async/await, batching visible-item captures and stabilizing wake-time polling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/stonerl/Thaw/pull/613?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Shared/Bridging/Bridging.swift
+++ b/Shared/Bridging/Bridging.swift
@@ -630,11 +630,16 @@ extension Bridging {
             return unionBounds
         }()
 
-        guard let display = content.displays.first(where: { $0.frame.intersects(effectiveBounds) })
-            ?? content.displays.first(where: { $0.frame.intersects(unionBounds) })
-            ?? content.displays.first
-        else {
-            diagLog.warning("captureWindowsImageSCK: no display matched bounds \(effectiveBounds)")
+        // Require a single display that fully contains BOTH the effective
+        // capture bounds and the union of all selected windows. Using
+        // intersects here would let a display that only partially overlaps
+        // through, producing a silently clipped capture; SCContentFilter
+        // (display:including:) clips at display.frame, so anything outside is
+        // lost without an error. A clean nil lets callers fall back or skip.
+        guard let display = content.displays.first(where: {
+            $0.frame.contains(effectiveBounds) && $0.frame.contains(unionBounds)
+        }) else {
+            diagLog.warning("captureWindowsImageSCK: no display fully contains effectiveBounds=\(effectiveBounds) and unionBounds=\(unionBounds)")
             return nil
         }
 
@@ -644,8 +649,10 @@ extension Bridging {
         configuration.showsCursor = false
         // boundsIgnoreFraming on the legacy API means "skip the window frame".
         // For a display+including filter the equivalent is ignoreShadowsDisplay;
-        // no per-window shadow toggle exists on this filter shape.
-        configuration.ignoreShadowsDisplay = options.contains(.boundsIgnoreFraming) || options.isEmpty
+        // no per-window shadow toggle exists on this filter shape. Empty
+        // options matches the legacy SkyLight default of keeping framing, so
+        // honor only the explicit flag here.
+        configuration.ignoreShadowsDisplay = options.contains(.boundsIgnoreFraming)
 
         let scale: CGFloat = options.contains(.nominalResolution)
             ? 1.0
--- a/Shared/Bridging/Bridging.swift
+++ b/Shared/Bridging/Bridging.swift
@@ -7,6 +7,7 @@
 //  Licensed under the GNU GPLv3
 
 import Cocoa
+@preconcurrency import ScreenCaptureKit
 
 // MARK: - Bridging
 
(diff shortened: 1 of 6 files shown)


diazdesandi:
.build

PR #636, opened 2026-05-28
Avoid persisting transitional always-hidden layout state

KaranocaVe:
## Summary
- avoid persisting saved layout snapshots while the hidden / always-hidden dividers are still transitioning
- block saved-layout reapply during the same divider-transition grace window
- add coverage for the new persistence gate behavior

## Bug
This bug is very subtle and still needs more real-world testing before we can know whether it is fully fixed.

Observed symptom:
- sometimes an icon gets moved into `always-hidden` for no obvious reason
- users experience it as icons being mysteriously or randomly pushed into the always-hidden section

Current hypothesis:
- a cache pass can observe the menu bar while the hidden / always-hidden control items are expanding or collapsing
- that transitional geometry can temporarily classify an item as `alwaysHidden`
- the transient classification then gets persisted into `savedSectionOrder`, and later layout restore applies it as if it were intentional

## Validation
- added a targeted unit test for the new section-divider transition persistence gate
- ran:
  - `xcodebuild test -quiet -scheme Thaw -destination 'platform=macOS' -derivedDataPath /tmp/thaw-codex-tests.577xm6 -only-testing:ThawTests/ShouldPersistSavedOrderTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=''`

## Risk
- this is a defensive fix around persistence and restore timing, not a proof that the underlying race can never happen in another path
- because the symptom is intermittent and timing-sensitive, more manual / long-running testing is still needed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved menu bar layout stability by preventing temporary state changes during menu section adjustments from being saved as persistent preferences.

[Review Change Stack](https://app.coderabbit.ai/change-stack/stonerl/Thaw/pull/636?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Thaw/MenuBar/ControlItem/ControlItem.swift
+++ b/Thaw/MenuBar/ControlItem/ControlItem.swift
@@ -48,7 +48,7 @@ final class ControlItem {
     }
 
     /// A hiding state for a control item.
-    enum HidingState {
+    enum HidingState: Equatable {
         case showSection
         case hideSection
     }
@@ -117,7 +117,14 @@ final class ControlItem {
     }
 
     /// The control item's hiding state (`@Published`).
-    @Published var state = HidingState.hideSection
+    @Published var state = HidingState.hideSection {
+        didSet {
+            guard oldValue != state, isSectionDivider else {
+                return
+            }
+            appState?.itemManager.recordSectionDividerTransition()
+        }
+    }
 
     /// The control item's window (`@Published`).
     @Published private(set) var window: NSWindow?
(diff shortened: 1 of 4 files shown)


nightah:
@KaranocaVe can you reproduce this behaviour you're suggesting needs fixing?
If so can you provide the diagnostic logs for said issue, along with an exaplanation as to what happened and what should have happened?

nightah:
Going to close this off, if you can come back and respond to my question and provide the appropriate detail happy to consider this, but until then, this PR doesn't really tell the team what it does and why we're implementing it.

PR #655, opened 2026-06-01
fix(resolution): stabilize Control Center widgets on a single display

### LINKED ISSUE #643: [Bug]: Little Snitch will not stick (beta 13 only)

### Description

Beta 12 worked fine, but beta 13 does not organize Little Snitch properly. After I move it into position, it reverts back to the location where I have 'New Items' set to appear. This is the same issue that Bartender had, so it seems to be using a similar method to organize menu bar items.

### Steps to Reproduce

put Little Snitch in place. Watch beta 13 go through every icon and misplace LS

### App Version

beta 13

### macOS Version

Tahoe

### Monitor Setup

Single Monitor

### Additional Information

_No response_

nightah:
## What does this PR do?

Fixes Little Snitch (and other Control Center-hosted menu bar widgets) being mis-identified as `com.apple.controlcenter:Item-0` and relocated on launch when only a single display is connected, by stopping the relocation of unresolved orphans and briefly creating a hidden virtual display to force the marker windows that source-PID resolution needs.

## PR Type

- [x] Bugfix
- [ ] CI/CD related changes
- [ ] Documentation
- [x] Feature
- [ ] Refactor
- [ ] Performance improvement
- [x] Test addition or update

- [ ] Yes
- [x] No

On macOS 26 every menu bar status item is hosted by Control Center, so `kCGWindowOwnerPID` and the AX owner point at Control Center rather than the creating app. The XPC `SourcePIDCache` recovers the real owner by pairing each icon with a bundle-ID-titled marker window, but the window server only publishes those markers when two or more displays are present. On a single display the markers never appear, so a widget whose status item Control Center hosts without its own `AXExtrasMenuBar` (Little Snitch's agent) stays unidentified as `com.apple.controlcenter:Item-0` and gets relocated by the layout pass, so it keeps moving on launch and only matches once the marker happens to surface much later.
Issue Number: #643

## What is the new behavior?

- Layer 1: `partitionUnmanagedUIDs` leaves a generic Control Center item with a nil source PID in place instead of treating it as unmanaged and moving it, so an unidentified widget never jumps around while resolution is pending.
- Virtual-display provoke: when exactly one display is connected and an orphan stays unresolved past a short grace, `VirtualDisplayProvoker` briefly creates a headless `CGVirtualDisplay` (bound through an Objective-C shim around the private API) so the window server publishes the marker windows; marker-pair then resolves the orphan to its real owner, typically within ~0.8s, and the display is torn down.
- Persistence and convergence: the resolved windowID-to-PID mapping persists in the XPC cache after teardown, and the provoker refreshes the manager's published item cache once resolution is confirmed, so the icon converges to its real identity immediately instead of waiting for an unrelated cache cycle. A 300s per-window cooldown and a 12s hold ceiling bound the mechanism.
- Containment: the phantom is excluded from Thaw's display enumeration via `Bridging.excludedDisplayID` and `NSScreen.managedScreens` so it never affects the Displays settings pane, overlay panels, or profile auto-switch, and its self-inflicted `didChangeScreenParametersNotification` is ignored by `DisplaySettingsManager` for the phantom's lifetime plus a short grace, so it no longer cancels the in-flight item-cache cycle and flashes a bar of orphans.

## PR Checklist

## Other information

Validated from field diagnostic logs on macOS 26.5 across multi-display and single-display reboots: on multi-display the provoke correctly stays dormant and marker-pair resolves Little Snitch naturally; on a true single display the provoke resolves Little Snitch to its real owner in ~0.8s, the resolution persists, the visible section converges to `at.obdev.littlesnitch.agent:Item-0`, and the item never moves. A ~24 minute single-display session showed zero orphan-flash blips and zero relocations.

### Notes for reviewers

- The provoke uses the private `CGVirtualDisplay` family via an Objective-C shim (`ThawVirtualDisplay.{h,m}`) wired in through `Thaw-Bridging-Header.h`; classes are resolved with `NSClassFromString` and guarded with `@try`/`@catch`, and `VirtualDisplay.isSupported` gates the whole feature so an absent class on a future OS degrades to a no-op rather than crashing.
- The hidden virtual display uses a hardcoded stable EDID so repeated provokes are not treated as a brand-new display each time.
- Known uncovered edge: if `CGVirtualDisplay` is unavailable (`isSupported == false`), the single-display case has no fallback; this was unobserved in the field, so no fallback was built.

Fixes #643

## Summary by CodeRabbit

* **New Features**
  * Added virtual-display support to temporarily create and exclude a transient display to aid menu-bar item resolution on single-display systems.

* **Improvements**
  * Display enumeration now ignores the transient virtual display and uses a managed-screen view across the app.
  * Menu-bar item cache now refreshes more frequently and exposes layout-stabilization state.
  * Adaptive background capture and per-display settings align with managed screens.

* **Tests**
  * New and extended tests for layout flattening, unmanaged-item partitioning, and profile-layout log replay.

### DIFF
--- a/Shared/Bridging/Bridging.swift
+++ b/Shared/Bridging/Bridging.swift
@@ -79,6 +79,17 @@ extension Bridging {
 extension Bridging {
     // MARK: Private Display Helpers
 
+    /// A display to exclude from Thaw's display enumeration while it exists.
+    ///
+    /// Set to the identifier of the transient virtual display that
+    /// VirtualDisplayProvoker creates to provoke marker-pair resolution on a
+    /// single-display machine, and cleared when it is removed. Filtering it
+    /// here keeps the phantom out of every display list derived from
+    /// getActiveDisplayList (including the active-menu-bar-display lookup),
+    /// so it never drives profile auto-switch or per-display state. nil during
+    /// normal operation, which makes the filter a no-op.
+    nonisolated(unsafe) static var excludedDisplayID: CGDirectDisplayID?
+
     private static func getActiveDisplayCount() -> UInt32? {
         var count: UInt32 = 0
         let result = CGGetActiveDisplayList(0, nil, &count)
@@ -98,6 +109,9 @@ extension Bridging {
         guard result == .success else {
             diagLog.error("CGGetActiveDisplayList failed with error \(result.logString)")
             return []
+        }
+        if let excluded = excludedDisplayID {
+            list.removeAll { $0 == excluded }
         }
         return list
     }
--- a/Shared/Bridging/Bridging.swift
+++ b/Shared/Bridging/Bridging.swift
@@ -7,6 +7,7 @@
 //  Licensed under the GNU GPLv3
 
 import Cocoa
+import os
 @preconcurrency import ScreenCaptureKit
 
 // MARK: - Bridging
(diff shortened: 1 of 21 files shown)


diazdesandi:
.build

stonerl (approved):

Issue #641, opened 2026-05-28
[Bug]: Every time my mac wake after sleep, Thaw reload itself, reloading every app that have an icon in menu bar causing duplicate of some apps


MilesTEG1:
### Search for Similar Reports

- [x] I have searched existing issues for similar reports

### Description

Hello,
Every time my mac wake after sleep, Thaw reload itself, reloading every app that have an icon in menu bar, and cause to have some duplicates in the menu bar (like BusyCal Menu & Alarms, or Little Snitch Agent...
Many windows are show in foreground, like Bitdefender window...

### Steps to Reproduce

1. Let the mac go to sleep
2. Wait some time
3. Wake up the mac

### App Version

2.0.0-beta.13

### macOS Version

26.5 (25F71)

### Monitor Setup

Single Monitor

### Additional Information

This abnormal behavior happens since beta 8-9 I think... But I'm not sure.
It may be related to another app, I suspect [AppHalt](https://apphalt.app/), but really not sure...

This problem happens when I'm on single monitor (the macbook display), or when I'm on dualscreen with a LG monitor.

itradesize:
same here, please fix it

stonerl:
In case you are on a MacBook with a notch can you please enable the option "Use LCS sorting on notched displays"

<img width="1133" height="856" alt="Image" src="https://github.com/user-attachments/assets/6f524955-5012-4fde-88e3-08d4aa14d6f8" />

itradesize:
Its a 16" macbook with notch ofc, but I'm only using my external screens. (3of them) 
Using it in clamshell mode.

gh0sti:
been happening to me as well the icons just keep reloading I had to disable thaw for the time being.

EDIT: @stonerl Just seeing your post I did enable the LCS sort and it seems to be stable for now. M5 MBA

MilesTEG1:
> In case you are on a MacBook with a notch can you please enable the option "Use LCS sorting on notched displays"
> 
[...]

PR #672, opened 2026-06-05
fix(resolution): resolve Control-Center-hosted items by title

nightah:
## What does this PR do?

Resolves the source app for menu bar widgets that host their `NSStatusItem` under Control Center while still publishing their own `AXExtrasMenuBar` child (AirBuddy, SpamSieve, Cotypist). These items were never resolved, kept a placeholder `com.apple.controlcenter` identity, and got swept into the hidden section on every profile re-layout. A new corroborated spatial fallback in `SourcePIDCache` attributes them to their real owner using a reverse-DNS title-to-bundle-id match, gated by corroboration rather than a looser distance threshold.

## PR Type

- [x] Bugfix
- [ ] CI/CD related changes
- [ ] Documentation
- [ ] Feature
- [ ] Refactor
- [ ] Performance improvement
- [x] Test addition or update

- [ ] Yes
- [x] No

N/A

A widget whose CG window is hosted by Control Center but whose own app publishes an `AXExtrasMenuBar` child (e.g. AirBuddy's `codes.rambo.AirBuddy.Menu`, owned by `codes.rambo.AirBuddyHelper`) fails source-PID resolution. The hosting CG slot is wider than the real icon, so the icon's AX child center sits more than 1pt from the CG window center (AirBuddy ~2pt, SpamSieve up to ~8pt), and the strict 1pt spatial pass rejects it. The item keeps the placeholder `com.apple.controlcenter` namespace, which no longer matches the profile entry saved under the resolved bundle id, so it is treated as unmanaged and `planUnmanagedPlacement` defaults it into the hidden section. The icon "goes into hiding" while pinned to the bar, and re-hides on every subsequent re-layout.
Issue Number: #641

## What is the new behavior?

A corroborated spatial fallback runs after the strict 1pt pass and before the marker-pair pass. For each still-unresolved window whose reverse-DNS title is in an owner relationship with a candidate app's bundle identifier (`HostedItemOwnership.titleIndicatesOwner`: at least two shared leading components, plus a full component-prefix or a first-differing-component prefix, case-insensitive), it accepts that app's nearest enabled extras-bar child within a 20pt backstop radius. The title corroboration, not the distance, is the guard, so a nearby unrelated neighbor is never mis-attributed and a same-vendor different-app pair (`pl.maketheweb.pixelsnap2` vs `pl.maketheweb.cleanshotx`) is rejected. Items handled here have their own AX child, so they resolve on the spatial pass, keep their saved profile section, and never reach the marker-pair fallback or the unresolved diagnostic. The strict 1pt matcher, the marker-pair gate, and the unrelated 10pt and 50pt matchers are unchanged.

## PR Checklist

## Other information

A pure threshold change cannot fix this safely: across 7,634 unresolved diagnostic samples from 63 captured logs (four users, six weeks) the correct and incorrect matches are interleaved by distance. AirBuddy's correct match is at 2.06pt, but `com.wireguard.macos` sits 2.00pt from an unrelated `app.updatest.Updatest` child, and `pl.maketheweb.pixelsnap2` sits 9pt from the different-app `pl.maketheweb.cleanshotx`. The component-aware owner rule was validated against that full dataset: it accepts only AirBuddy, SpamSieve, and Cotypist and rejects every wrong neighbor at any distance. `HostedItemOwnership` lives in the shared module so both the XPC service and the test target reach it without adding a file; `HostedItemOwnershipTests` locks in the accept/reject cases drawn directly from the logs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved menu bar item detection and resolution, adding a spatial fallback to better identify Control Center–hosted items and reduce misattributions
  * Enhanced proximity and title-based ownership checks to prevent incorrect associations between items and apps

* **Tests**
  * Added comprehensive tests covering hosted-item ownership and detection edge cases

### DIFF
--- a/MenuBarItemService/SourcePIDCache.swift
+++ b/MenuBarItemService/SourcePIDCache.swift
@@ -393,6 +393,53 @@ final class SourcePIDCache {
             }
         }
 
+        // Corroborated spatial fallback for Control-Center-hosted items
+        // whose own app DOES publish an extras-bar AX child, but offset from
+        // the CG window center by more than the strict 1pt pass tolerates.
+        // The hosting CG slot is wider than the real icon, so their centers
+        // diverge: AirBuddy's by ~2pt, SpamSieve's by up to ~8pt. Accept the
+        // nearest such child within a generous radius ONLY when the window's
+        // reverse-DNS title is in an owner relationship with the app's bundle
+        // identifier (HostedItemOwnership). The title corroboration, not the
+        // distance, is what makes this safe: a nearby unrelated neighbor
+        // (WireGuard's slot beside Updatest at ~2pt) fails the owner check and
+        // is left for later passes. Runs BEFORE marker-pair so items that have
+        // their own AX child are claimed here and never reach that fallback.
+        // Empirically the furthest correct owner-corroborated match across
+        // captured logs is ~15pt; 20 leaves margin while staying well inside
+        // a neighbor's slot. The owner check is the real guard.
+        let hostedExtrasMatchRadius: CGFloat = 20
+        for app in apps {
+            if unresolvedWindows.isEmpty { break }
+            guard let appBundleID = app.bundleIdentifier else { continue }
+            let candidateWindows = allWindows.filter {
+                unresolvedWindows.contains($0.windowID)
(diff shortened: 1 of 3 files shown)


stonerl (approved):
LGTM

PR #677, opened 2026-06-06
fix(menubar): defer layout while a relaunched app's item re-pairs

nightah:
## What does this PR do?

Stops previously-hidden menu bar items from jumping back onto the bar when an app with a status item relaunches (e.g. an in-app update like Free Download Manager). It arms a short settling period on a tracked app's relaunch so the layout move pass and the virtual-display provoke both wait until the item has re-paired, then plan against stable geometry.

## PR Type

- [x] Bugfix
- [ ] CI/CD related changes
- [ ] Documentation
- [ ] Feature
- [ ] Refactor
- [ ] Performance improvement
- [x] Test addition or update

- [ ] Yes
- [x] No

N/A

When an app with a menu bar item relaunches, its status item disappears and re-registers over a few seconds, during which the bar is in flux: the Thaw hidden control item (the wide boundary window) is briefly unresolved and its position slides hundreds of points between cycles. The `didLaunchApplicationNotification` handler refreshes the cache immediately, which dispatches a bulk apply, and the move pass runs against that transient geometry. The planner's section decision is actually correct (`crossSectionMoves=0, totalSectionMismatch=0` in the field log), but the within-section reorder moves resolve their anchors against the sliding control item and land previously-hidden items next to a visible Control Center item, so hidden icons jump back onto the bar. The same relaunch churn also fires the virtual-display provoke prematurely for the transient orphan.
Issue Number: #641

## What is the new behavior?

`isSettling` already makes both the move pass (`applyProfileLayout` awaits `waitForStartupSettlingToEnd`) and the provoke (guarded by `isSettling`) hold off while the bar is unstable, but it was only ever started for cold boot and Thaw-initiated relaunch waves, never for a spontaneous external relaunch. The launch handler now starts a settling period, but only when the launched bundle ID owns a menu bar item Thaw already tracks, so ordinary launches (apps with no status item) arm nothing and pay no deferral. The period uses expected-set mode keyed on that bundle ID, exiting the instant the item re-pairs with a resolved PID (median ~3s across field logs, p95 ~6s); the 8s `maxDuration` is only a backstop. One change covers both the bulk-apply unhide and the premature provoke. The tracked-app gate is extracted into `MenuBarItemManager.tracksMenuBarItem(bundleID:in:)` so it is unit-testable.

## PR Checklist

## Other information

The unhide is an execution-time effect (anchor resolution against transient geometry), not a planner-decision error, the planner logged `totalSectionMismatch=0` at the moment items unhid. That is why the fix defers the apply rather than changing the planner, and why the log-replay harness (which models the pure planner decision, correct in both transient and settled states) cannot reproduce this class; an FDM fixture there would pass with or without the fix, so none was added. `RelaunchSettlingGateTests` is a red/green guard on the gate condition (including the `org.x.fdm6` vs `org.x.fdm6x` boundary), which is the precise condition under which the unhide is prevented.

## Summary by CodeRabbit

* **Improvements**
  * More robust handling of app launches for tracked menu bar items: uses the launched app's bundle identifier to start a short, bundle-scoped startup settling period, reducing spurious refreshes.

* **Tests**
  * Added tests validating tracked vs untracked bundle identifiers, namespace/format edge cases, and empty-set behavior.

### DIFF
--- a/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
+++ b/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
@@ -1230,9 +1230,35 @@ final class MenuBarItemManager: ObservableObject {
             for: NSWorkspace.didLaunchApplicationNotification
         )
         .debounce(for: 1, scheduler: DispatchQueue.main)
-        .sink { [weak self] _ in
+        .sink { [weak self] notification in
             guard let self else { return }
-            MenuBarItemManager.diagLog.debug("App launched, refreshing cache for potential new items")
+            let launchedBundleID = (notification.userInfo?[NSWorkspace.applicationUserInfoKey] as? NSRunningApplication)?
+                .bundleIdentifier
+            MenuBarItemManager.diagLog.debug(
+                "App launched\(launchedBundleID.map { " (\($0))" } ?? ""), refreshing cache for potential new items"
+            )
+
+            // If the launched app is one we already track a menu bar item for,
+            // it just relaunched (e.g. an in-app update): its status item is
+            // about to disappear and re-register, churning the bar for a few
+            // seconds. Start a settling period keyed on its bundle ID so the
+            // move pass (applyProfileLayout waits on waitForStartupSettlingToEnd)
+            // and the virtual-display provoke (guarded by isSettling) both hold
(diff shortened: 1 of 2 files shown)


nightah:
.build

stonerl (approved):

PR #691, opened 2026-06-09
feat(displays): warn before spacing relaunches, save choice per profile

### LINKED ISSUE #685: Display connect/disconnect/switch can repeatedly relaunch menu bar apps without runtime confirmation
# Display connect/disconnect/switch can repeatedly relaunch menu bar apps without runtime confirmation

Follow-up to #591. I am opening this as a separate issue because the maintainer asked to track behavior changes separately from the original closed thread.

## Summary

Thaw can relaunch apps with menu bar items during ordinary display connect/disconnect/switch flows when the spacing configured for the active menu bar display differs from the currently applied system spacing.

The surprising part for me was not that pressing “Apply” changes spacing. I understood that part. The surprising part was that after different spacing is saved for displays/templates, Thaw can re-apply spacing on later display connect/disconnect/switch events and relaunch apps again without asking at that moment.

As a user, I would expect a spacing change to take effect either at the moment I explicitly press Apply, or perhaps after a system restart. Auto-applying it on later display connect/disconnect/switch events is not obvious.

From reading the current `development` code and `2.0.0-beta.14`, the runtime display-change path can apply the active display spacing after `screenParametersChanged`, and that path does not appear to show a just-in-time confirmation or offer a way to postpone the relaunch.

In my case this caused unrelated apps to be relaunched unexpectedly, and I lost typed input in some applications.

## Why this is serious

Connecting, disconnecting, or switching displays is a routine user action. A user does not naturally expect it to terminate and reopen unrelated apps.

Even if the relaunch is technically needed to apply `NSStatusItemSpacing`, it is still a destructive side effect from the user's point of view: apps can lose unsaved input, progress, or transient state. The user may be typing in another app when the display state changes.

This also makes the feature surprising because the UI presents spacing as a per-display customization, while the underlying macOS setting is effectively system-wide.

It is especially surprising because the Global spacing/template value is saved as soon as the slider is changed, even before “Apply to All Displays” is pressed. It does not immediately apply the spacing or relaunch apps, but that saved template can later be used when displays are seeded/applied. From the user's point of view, this means a value can be saved first and only become destructive later during a display state change.

That creates a dangerous combination: a setting can be saved without an explicit call to action and confirmation popup at the moment it is saved; later, it can be applied automatically during display changes; and the user may not even understand which app is causing unrelated apps to restart.

## Code path I am concerned about

My understanding from the current code, with AI assistance:

- `DisplayIceBarConfiguration.itemSpacingOffset` notes that macOS reads `NSStatusItemSpacing` as a single system-wide value, so Thaw writes that value and relaunches apps whenever a display becomes or remains the active menu bar display.
- The per-display spacing slider appears to be draft-only until `Apply`, but the Global spacing slider writes to `displaySettings.globalConfiguration` immediately.
- `DisplaySettingsManager` persists `globalConfiguration`, and newly detected displays are seeded from the current global template.
- `DisplaySettingsManager` listens for `didChangeScreenParametersNotification`. After display changes, if the active menu bar display changed, it calls `applyActiveDisplaySpacing(reason: "screenParametersChanged")`.
- `applyActiveDisplaySpacing` reads the active display's `itemSpacingOffset`, assigns it to `spacingManager.offset`, and calls `applyOffset()`.
- `MenuBarItemSpacingManager.applyOffset()` writes the defaults and relaunches affected menu bar item apps when the on-disk spacing does not already match the desired spacing.

So if display changes keep moving between displays/templates with different desired spacing values, the relaunch wave can happen each time Thaw needs to switch the system spacing, as part of runtime display-change handling rather than as an explicit Settings action.

## Expected behavior

Thaw should avoid silently relaunching unrelated apps during routine display state changes.

Possible mitigations:

- Use unified spacing across displays as the safe default.
- Treat per-display spacing as an explicit advanced/dangerous opt-in.
- If a display connect/disconnect/switch would require relaunching apps, ask the user at that moment.
- Offer choices such as:
  - relaunch now;
  - postpone;
  - keep current spacing for now;
  - disable per-display spacing / use unified spacing.

## Actual behavior

The current warning added in #670 improves disclosure, but the runtime behavior can still relaunch apps without a just-in-time confirmation each time a display state change requires a different spacing value.

## Related

- #591
- #670
- https://github.com/stonerl/Thaw/issues/686

### LINKED ISSUE #686: Make display-spacing relaunch warnings mention repeated display events, each affected app, and data-loss risk
# Make display-spacing relaunch warnings mention repeated display events, each affected app, and data-loss risk

Follow-up to #591 and to the behavior-level issue: https://github.com/stonerl/Thaw/issues/685.

I am opening this separately because the maintainer asked to track follow-up improvements separately, and because this is narrower than changing the runtime relaunch behavior itself.

## Summary

#670 added a persistent warning in Displays settings:

> Connecting a display whose spacing differs from the current spacing relaunches every menu bar app. Keep spacing consistent across displays to avoid relaunches.

That is a useful improvement, but I think the warning copy still understates the practical impact. The user needs to understand that this can happen on each later display connect/disconnect/switch event while spacing differs, that Thaw may terminate and reopen each app with a menu bar item, and that this can cause unsaved input, progress, or transient app state to be lost.

This happened in my case: I lost typed input in some applications after an unexpected relaunch wave.

## Suggested persistent warning text

> Each time connecting, disconnecting, or switching displays makes Thaw apply a different menu bar spacing, Thaw relaunches each app with a menu bar item. Keep spacing consistent across displays to avoid repeated relaunches. Relaunching apps may cause unsaved input, progress, or transient app state to be lost.

Why these changes:

- `Each time` is important because the surprise is not only the first manual Apply. The relaunch can happen again on later display connect/disconnect/switch events while spacing differs.
- `connecting, disconnecting, or switching displays` is clearer because the risk is tied to routine display state changes, not only an explicit settings action.
- `each app with a menu bar item` is more concrete than “menu bar app”.
- Mentioning possible loss of unsaved input/progress/state explains the real user impact instead of only describing the technical relaunch.

## Other warning text that should be strengthened

The same data-loss wording should be used consistently in the related alerts and annotations.

For example, instead of:

> Applying this spacing change will briefly relaunch all apps with menu bar items.

Consider:

> Applying this spacing change will relaunch each app with a menu bar item. Relaunching apps may cause unsaved input, progress, or transient app state to be lost.

For non-active displays, the warning should also explain the delayed risk:

> Saving this spacing for a non-active display may relaunch apps later when that display becomes the active menu bar display.

For “Apply global settings to all displays?”, the text should avoid the weaker “may briefly relaunch apps” phrasing and explicitly state the impact:

> This will overwrite the settings of N displays with the global template. If the active display's spacing changes, Thaw will relaunch each app with a menu bar item. Relaunching apps may cause unsaved input, progress, or transient app state to be lost.

## Visual severity

Because this can lead to loss of unsaved user work, I think this should be styled as a danger/critical warning rather than a normal warning. The risk is closer to a destructive operation than a cosmetic preference.

## Acceptance criteria

- The persistent display-spacing warning mentions repeated connect/disconnect/switch events and possible unsaved input/progress/state loss.
- Apply/global confirmation alerts mention that each affected app can be relaunched.
- Non-active display spacing edits warn that the relaunch may happen later when that display becomes active.
- The warning uses a visual treatment strong enough for possible user-work loss.

## Related

- #591
- #670
- https://github.com/stonerl/Thaw/issues/685

nightah:
## What does this PR do?

Adds a just-in-time confirmation before menu bar spacing changes relaunch your apps, an override toggle to disable it, and persists the new confirmation settings per profile. Also rewords the existing relaunch warnings to state the data-loss risk plainly.

## PR Type

- [ ] Bugfix
- [ ] CI/CD related changes
- [ ] Documentation
- [x] Feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] Test addition or update

- [ ] Yes
- [x] No

Profiles saved before this change decode with backward-compatible defaults (`decodeIfPresent`), so existing profiles load unchanged.

Changing menu bar item spacing relaunches every app that owns a menu bar item, which can lose unsaved input, in-progress work, or transient app state. The automatic display-transition path applies a different spacing and fires this relaunch wave with no warning at all, and the in-pane Apply and global-broadcast confirmations describe the relaunch as a brief, harmless event.
Issue Number: #685, #686

## What is the new behavior?

When a display transition would actually rewrite the on-disk spacing, `DisplaySettingsManager.applyActiveDisplaySpacing` now presents an app-modal `NSAlert` before relaunching, so it surfaces even with the Settings window closed. The prompt is gated by a new `MenuBarItemSpacingManager.willRelaunch(forOffset:)` predicate that mirrors the existing no-op guard, so it never appears for a no-op; declining keeps the current on-disk spacing and leaves `lastAppliedActiveDisplayUUID` untouched so the next genuine transition re-prompts.
A new `confirmSpacingRelaunch` toggle (default on) in the Displays pane governs all spacing confirmations: when off, the automatic path applies silently and the in-pane Apply and global-broadcast paths bypass their alerts. A companion `unconfirmedSpacingProfileScope` picker chooses whether an unconfirmed apply saves spacing to the active profile only or to every profile when a profile is active, and the alert carries a "Don't ask again" suppression checkbox wired to the toggle.
Both new settings are full profile state: captured in `saveProfile` and `applyCurrentConfiguration`, restored in `applySnapshot`, threaded through `Profile`/`ProfileContent`, and reset in `resetDisplay`, so each profile can carry its own confirmation behaviour. The three existing relaunch warning strings are reworded to state the data-loss risk plainly, and the new UI strings are added to the catalog.

## PR Checklist

Closes #685, Closes #686

## Summary by CodeRabbit

* **New Features**
  * Confirmation dialog when spacing changes would relaunch apps
  * Toggle to disable relaunch confirmations for faster workflow
  * Option to choose where unconfirmed spacing edits are saved (active profile or all profiles)

* **Improvements**
  * Clearer warning text about per-app relaunch and possible loss of unsaved/transient state
  * UI flows now respect the confirmation setting when applying or broadcasting spacing changes

### DIFF
--- a/Thaw/Settings/Models/DisplaySettingsManager.swift
+++ b/Thaw/Settings/Models/DisplaySettingsManager.swift
@@ -34,6 +34,17 @@ final class DisplaySettingsManager: ObservableObject {
     /// re-connect the display first.
     @Published var knownDisplays: [String: KnownDisplay] = [:]
 
+    /// Whether Thaw asks for confirmation before a spacing change relaunches
+    /// menu bar apps. When true, the automatic display-transition path shows
+    /// a just-in-time prompt and the Displays pane shows its Apply/global
+    /// confirmation alerts. When false, both apply without asking.
+    @Published var confirmSpacingRelaunch = Defaults.DefaultValue.confirmSpacingRelaunch
+
+    /// When confirmSpacingRelaunch is off and a profile is active, selects
+    /// whether an applied spacing change is saved to the active profile only
+    /// or to every profile.
+    @Published var unconfirmedSpacingProfileScope = Defaults.DefaultValue.unconfirmedSpacingProfileScope
+
     /// Storage for internal observers.
     private var cancellables = Set<AnyCancellable>()
 
@@ -168,6 +179,12 @@ final class DisplaySettingsManager: ObservableObject {
             } catch {
                 diagLog.error("Failed to decode known display cache: \(error)")
             }
+        }
+        Defaults.ifPresent(key: .confirmSpacingRelaunch, assign: &confirmSpacingRelaunch)
+        if let raw = Defaults.string(forKey: .unconfirmedSpacingProfileScope),
+           let scope = SpacingProfileSaveScope(rawValue: raw)
+        {
+            unconfirmedSpacingProfileScope = scope
         }
     }
 
@@ -338,6 +355,13 @@ final class DisplaySettingsManager: ObservableObject {
             }
             .store(in: &c)
 
+        $confirmSpacingRelaunch.persistToDefaults(key: .confirmSpacingRelaunch, in: &c)
+        $unconfirmedSpacingProfileScope.persistToDefaults(
+            key: .unconfirmedSpacingProfileScope,
+            transform: \.rawValue,
+            in: &c
+        )
+
         cancellables = c
     }
 
@@ -366,8 +390,23 @@ final class DisplaySettingsManager: ObservableObject {
     /// moving them.
     private func applyActiveDisplaySpacing(reason: String) {
         guard let appState else { return }
-        lastAppliedActiveDisplayUUID = Bridging.getActiveMenuBarDisplayUUID()
         let desired = Int(configurationForActiveDisplay().itemSpacingOffset.rounded())
+        // A display transition can fire the relaunch wave with no warning.
+        // When confirmations are enabled and this apply would actually
+        // relaunch apps, ask the user first. Declining keeps the current
+        // on-disk spacing and leaves lastAppliedActiveDisplayUUID untouched
+        // so the next genuine transition re-prompts. The in-pane Apply and
+        // global broadcast carry their own confirmations, so only the
+        // automatic path is gated here.
+        if reason == "screenParametersChanged",
+           confirmSpacingRelaunch,
+           appState.spacingManager.willRelaunch(forOffset: desired),
+           !presentSpacingRelaunchConfirmation()
+        {
+            diagLog.info("User declined the spacing relaunch confirmation for a display transition; skipping apply")
+            return
+        }
+        lastAppliedActiveDisplayUUID = Bridging.getActiveMenuBarDisplayUUID()
         appState.spacingManager.offset = desired
         Task { [weak self] in
             guard let self else { return }
@@ -401,6 +440,27 @@ final class DisplaySettingsManager: ObservableObject {
                 diagLog.error("applyActiveDisplaySpacing(\(reason)) failed: \(error)")
             }
         }
+    }
+
+    /// Presents an app-modal confirmation before a display transition fires
+    /// the relaunch wave. Returns true when the user approves the relaunch,
+    /// false when they cancel. Runs modally so it surfaces even with the
+    /// Settings window closed; ticking the suppression checkbox turns
+    /// confirmSpacingRelaunch off so future transitions apply silently.
+    private func presentSpacingRelaunchConfirmation() -> Bool {
+        NSApp.activate(ignoringOtherApps: true)
+        let alert = NSAlert()
+        alert.messageText = String(localized: "Apply menu bar spacing change?")
+        alert.informativeText = String(localized: "When a display transition requires Thaw to apply a different menu bar spacing, Thaw relaunches apps with menu bar items. Relaunching apps may cause unsaved input, progress, or transient app state to be lost.")
+        alert.addButton(withTitle: String(localized: "Apply"))
+        alert.addButton(withTitle: String(localized: "Cancel"))
+        alert.showsSuppressionButton = true
+        alert.suppressionButton?.title = String(localized: "Don't ask again")
+        let response = alert.runModal()
+        if alert.suppressionButton?.state == .on {
+            confirmSpacingRelaunch = false
+        }
+        return response == .alertFirstButtonReturn
     }
 
     /// Handles per-display settings changed externally via Settings URI scheme.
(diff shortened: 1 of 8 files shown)


Thaw/Settings/Models/DisplaySettingsManager.swift
```diff
@@ -403,6 +442,27 @@ final class DisplaySettingsManager: ObservableObject {
...
+        let alert = NSAlert()
+        alert.messageText = String(localized: "Apply menu bar spacing change?")
+        alert.informativeText = String(localized: "When a display transition requires Thaw to apply a different menu bar spacing, Thaw relaunches apps with menu bar items. Relaunching apps may cause unsaved input, progress, or transient app state to be lost.")
+        alert.addButton(withTitle: String(localized: "Apply"))
+        alert.addButton(withTitle: String(localized: "Cancel"))
+        alert.showsSuppressionButton = true
+        alert.suppressionButton?.title = String(localized: "Don't ask again")
+        let response = alert.runModal()
+        if alert.suppressionButton?.state == .on {
+            confirmSpacingRelaunch = false
```
fleytman, line 462: One UX concern here: the suppression checkbox is applied independently of the modal response. If a user checks “Don't ask again” and then presses `Cancel`, the current apply is cancelled, but future confirmations are disabled, so a later display transition can apply spacing and relaunch apps silently.

That feels like remembering the checkbox rather than the action the user chose. Could this either only honor suppression when the user presses `Apply`, or make the checkbox consequence explicit, e.g. “Don't ask again and apply future spacing changes without confirmation”?

fleytman (approved):

stonerl (approved):

PR #704, opened 2026-06-12
fix(window-info): don't warn on empty window identifier lists

HAOGRE:
## What does this PR do?

Stops `WindowInfo.createWindows(from:)` from logging a warning when it is
called with an empty window-identifier list. An empty list is a valid,
expected input — there are simply no windows to describe — so it should
return an empty array silently.

## PR Type

- [x] Bugfix

- [x] No

`Bridging.createCGWindowArray(with:)` returns `nil` by design for an empty
identifier list. `WindowInfo.createWindows(from:)` treats that `nil` as a
failure and logs:

```
createWindows: createCGWindowArray returned nil for 0 window IDs
```

On macOS 27 the per-item menu bar window list is frequently empty (Apple
unified all menu bar items into a single window owned by the new
`MenuBarAgent` process), so menu-bar enumeration paths repeatedly call
`createWindows(from: [])` and flood the diagnostic log with this warning
many times per minute. That noise makes the diagnostic logs maintainers
ask users to attach much harder to read.

Issue Number: #687

## What is the new behavior?

`createWindows(from:)` short-circuits on an empty input and returns an
empty array without logging. Genuine failures (a non-empty list that still
yields `nil`) keep their warning. A regression test covers both the return
value and the absence of the warning for empty input.

## PR Checklist

## Other information

This is a small, self-contained noise-reduction fix scoped to log
hygiene. It does not attempt full macOS 27 menu bar management (tracked in
#687 and the maintainers' `feat/macos-27-experimental` branch); it only
keeps the diagnostic log readable on macOS 27 so that effort is easier to
debug.

### Notes for reviewers

The new test (`WindowInfoCreateWindowsTests`) enables the diagnostic
logger to a temporary file, calls `createWindows(from: [])`, then writes a
sentinel line and waits for it to flush (the logger's write queue is
serial/FIFO) before asserting the empty-input warning is absent. It fails
on the previous behavior and passes with this change.

### DIFF
--- a/Shared/Utilities/WindowInfo.swift
+++ b/Shared/Utilities/WindowInfo.swift
@@ -106,6 +106,15 @@ extension WindowInfo {
     ///
     /// - Parameter windowIDs: A list of window identifiers.
     static func createWindows(from windowIDs: [CGWindowID]) -> [WindowInfo] {
+        // An empty identifier list is a valid, expected input — there are
+        // simply no windows to describe. This happens routinely on macOS 26
+        // (Tahoe) and especially on macOS 27, where the per-item menu bar
+        // window list is frequently empty. `createCGWindowArray` returns nil
+        // by design for an empty list, so return early to avoid logging a
+        // misleading warning on every such call.
+        guard !windowIDs.isEmpty else {
+            return []
+        }
         guard let array = Bridging.createCGWindowArray(with: windowIDs) else {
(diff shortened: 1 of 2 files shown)


diazdesandi:
The only reason this would become spammy is if the app is completely broken, like in earlier macOS 27 builds, where the underlying issue is much bigger than having extra warnings. With the current macOS 27 experimental branch/build there are no issues with log spam from this warning, so I don’t think it needs to be suppressed.

PR #817, opened 2026-07-26
Resolve small layout divergences with targeted moves

alvst:
## Scope

## Change

Detects up to two divergent menu-bar items and moves them directly to their saved section boundary. Larger or unsafe divergences continue to use the existing full-profile fallback.

## Why

Avoids an unnecessary full layout reapply when only a small number of items have drifted.

## Validation


diazdesandi:
Two blockers:

1. **Branch noise.** This forked with #761–#765 still on it, so GitHub shows ~50 files. The real change is just `3188fca0` (one file in `MenuBarItemManager`). Please rebase onto latest `integration/menubar-scramble` and drop the unrelated commits.

2. **Design.** `applyTargetedSectionMoves` parks items at the section edge, not their saved index. The divergence check only looks at section membership, so wrong order inside a section never gets fixed afterward.

#822 attacks the same “transient widget → full notched re-drag” symptom by trimming the full-sort replay. I’d like to land that first and see if this is still needed. If it is, please redo it so moves go to the **saved index**, with planning as a pure `LayoutSolver` helper that’s easy to test.

PR #818, opened 2026-07-26
Avoid spurious saved-layout bulk applies on recache

alvst:
## Scope

## Change

Gates saved-layout reapplication on actual window-ID changes rather than every recache cycle, with focused coverage for that gate.

## Why

Prevents repeated bulk layout work when the underlying menu-bar item set has not changed.

## Validation

### DIFF
--- a/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
+++ b/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
@@ -6811,6 +6811,28 @@ extension MenuBarItemManager {
         return !previous.isSubset(of: current)
     }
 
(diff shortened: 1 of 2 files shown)


diazdesandi:
Closing as already addressed on `integration/menubar-scramble`.

Scramble already gates saved-layout reapplication with `windowIDsChanged || layoutDiverged` in `MenuBarItemManager` (see the saved-layout gate around that check). Retargeting this PR would mostly conflict in favor of scramble’s version, so we’re not carrying it forward as a separate change.

Thanks for isolating the fix — the intent is covered on scramble.

diazdesandi:
Superseded by `integration/menubar-scramble`’s `windowIDsChanged || layoutDiverged` saved-layout gate.

PR #819, opened 2026-07-26
Restore cursor visibility across move retries

alvst:
## Scope

## Change

Moves cursor-hide ownership to the outer retry operation so every retry path restores the cursor and the watchdog consistently covers the whole operation.

## Why

Prevents the cursor from remaining hidden or being restored inconsistently after retried menu-bar moves.

## Validation

### DIFF
--- a/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
+++ b/Thaw/MenuBar/MenuBarItems/MenuBarItemManager.swift
@@ -2801,11 +2801,16 @@ extension MenuBarItemManager {
         event: CGEvent,
         item: MenuBarItem,
         timeout: Duration,
-        repeating count: Int
+        repeating count: Int,
+        manageCursorVisibility: Bool = true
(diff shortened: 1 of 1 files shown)


diazdesandi:
Closing as already addressed on `integration/menubar-scramble`.

Cursor hide/restore across bulk moves and retries landed there as:

- `c1311d72` — fix(cursor): stop bulk-apply loops from hijacking and hiding the cursor
- `9c77fb65` — fix(menubar): suppress per-item cursor warps during bulk applies
- `68bdec01` — fix(cursor): release the pointer as soon as the last move lands

Scramble’s approach also covers a race this PR’s outer-ownership shape doesn’t. Retargeting would largely conflict into scramble’s version, so we’re closing instead of rebasing.

Thanks for the isolation work.

diazdesandi:
Superseded on `integration/menubar-scramble` by c1311d72 / 9c77fb65 / 68bdec01.

PR #820, opened 2026-07-26
fix(menubar): avoid repeated source PID cache scans

alvst:
# Summary

Avoids source-PID lookups for Thaw control windows, temporarily negative-caches unresolved window IDs, and rate-limits repeated unresolved diagnostics.

**Scope:** One focused reliability fix for menu-bar window PID resolution. Broader 2.1 feature work remains parked for a later integration branch.

> We are on the process of migrating from XCTest to Swift Test. If you are adding new tests, please use Swift Test.
> **External contributors:** before opening a PR for a bug fix or new feature, please make sure there's a corresponding issue in the [issue tracker](https://github.com/thaw-app/Thaw/issues). PRs that fix or change things that haven't been reported/agreed on may be closed without review.

Closes: N/A

## PR Type

Describe **what this change does** (not the linked issue’s request kind). Bug *reports* use `bug` on issues; bug *fixes* use `fix` on PRs.

If you tick **Feature** or **Refactor** and touch more than ~20 files, please mention why this can’t be split.

- [x] Bug fix
- [ ] CI/CD
- [ ] Documentation
- [ ] Feature
- [ ] Enhancement
- [ ] Performance improvement
- [ ] Refactor
- [ ] Test addition or update
- [ ] Other (please describe)

## Area

**Product surfaces** (optional when the change is not about the app UI). Path-based labeling also applies.

- Use **PR Type** for *what* changed (`CI/CD`, `Documentation`, `Other` / chore, etc.).
- Use **`ops`** for *where* when it is repo operations: CI, release, GitHub hygiene, scripts, lint/sonar config — not a product surface.

- [x] menubar
- [ ] icebar
- [ ] layout
- [ ] appearance
- [ ] settings
- [ ] onboarding
- [ ] permissions
- [ ] profiles
- [ ] hotkeys
- [ ] updates
- [ ] ops

## Does this PR introduce a breaking change?

- [ ] Yes - if yes, please describe the impact and migration path
- [x] No

## What is the new behavior?

Reduces repeated expensive scans for window IDs that cannot currently be resolved by skipping source-PID lookups for Thaw control windows, negative-caching unresolved IDs briefly, and rate-limiting repeated unresolved diagnostics.

## PR Checklist

- [x] I've built and run the app locally and verified that it works as expected.
- [ ] I've run `swiftformat .` to keep the code style consistent.
- [ ] I've run the smallest relevant test commands (list 1–2 below), e.g. `xcodebuild test …` or `swift test --package-path MenuBarModel`.
[...]

Issue #795, opened 2026-07-17
Parked (off-screen) items whose title is their bundle identifier are unresolvable — one-way door into the hidden section


lathe-agent-oa:
## Summary

`SourcePIDCache` resolves menu bar item ownership through spatial passes (AX-frame proximity to the CG window) and the marker-pair pass, which considers **on-screen icons only**. A widget whose status window is titled with its own bundle identifier — Little Snitch's agent (`at.obdev.littlesnitch.agent`), already noted as a problem widget in the marker-pair comments — therefore resolves fine while visible, but becomes **unresolvable the moment it is parked at off-screen coordinates** in the hidden section. An unresolvable hidden item can never be matched back to its saved section, so one churn event that lands such an item in hidden is a one-way door: nothing can ever restore it.

## Observed (diagnostic log, exact strings)

Little Snitch parked in hidden, unresolvable — its own AX element is absent from the candidate list entirely, and the nearest candidate is another app ~1000pt away:

```
[DEBUG] [SourcePIDCache] SourcePIDCache diag unresolved: windowID=259 title=at.obdev.littlesnitch.agent bounds=(-5724.0, 0.0, 68.0, 33.0) center=(-5690.0, 16.5) | cgOwner=com.apple.controlcenter:pid=1156 ownerName=Control Center | closestAXFrame=(-9184.0, 3.0, 5002.0, 24.0) in app=com.stonerl.Thaw distance=993.0011329298673 closestAXEnabled=false | nearest=[com.stonerl.Thaw@993.0(enabled=false), com.knollsoft.Rectangle@1534.0(enabled=true), net.matthewpalmer.Rocket@1570.0(enabled=true)]
```

Sequence that produced it: an app-churn window (Stats widget reconfiguration + a Thaw relaunch) briefly broke the item's resolution; the layout apply could not match it to its saved visible slot; it was filed into hidden; parked there, every subsequent resolution attempt failed (16 consecutive `diag unresolved` events), so no divergence-triggered apply could ever pull it back out.

## Root-cause hypothesis

- **OBSERVED:** all resolution passes require geometry (spatial AX match) or on-screen presence (marker-pair); a parked item satisfying neither stays unresolved indefinitely.
- **INFERRED:** any widget in the marker-pair class (title = own bundle id, no AX extras child tracking the parked position) is susceptible; Little Snitch is the observed instance.

## Suggested fix

[...]

PR #856, opened 2026-07-30
fix(menubar): back off negative-cache TTL instead of flat 60s

lathe-agent-oa:
# Summary

Replace the flat 60s SourcePID negative-cache TTL from #820 with a per-window backoff ladder (5s → 15s → 60s) so a low-yield cold scan after launch does not permanently bar unresolved windows for the rest of the session.

**Scope:** Focused fix in `SourcePIDCache` plus a pure policy helper in `Shared/` and unit tests. No product UI changes.

> We are on the process of migrating from XCTest to Swift Test. If you are adding new tests, please use Swift Test.
> **External contributors:** before opening a PR for a bug fix or new feature, please make sure there's a corresponding issue in the [issue tracker](https://github.com/thaw-app/Thaw/issues). PRs that fix or change things that haven't been reported/agreed on may be closed without review.

Closes: N/A

Related to #820

## PR Type

Describe **what this change does** (not the linked issue’s request kind). Bug *reports* use `bug` on issues; bug *fixes* use `fix` on PRs.

If you tick **Feature** or **Refactor** and touch more than ~20 files, please mention why this can’t be split.

- [x] Bug fix
- [ ] CI/CD
- [ ] Documentation
- [ ] Feature
- [ ] Enhancement
- [ ] Performance improvement
- [ ] Refactor
- [x] Test addition or update
- [ ] Other (please describe)

## Area

**Product surfaces** (optional when the change is not about the app UI). Path-based labeling also applies.

- Use **PR Type** for *what* changed (`CI/CD`, `Documentation`, `Other` / chore, etc.).
- Use **`ops`** for *where* when it is repo operations: CI, release, GitHub hygiene, scripts, lint/sonar config — not a product surface.

- [x] menubar
- [ ] icebar
- [x] layout
- [ ] appearance
- [ ] settings
- [ ] onboarding
- [ ] permissions
- [ ] profiles
- [ ] hotkeys
- [ ] updates
- [ ] ops

- [ ] Yes - if yes, please describe the impact and migration path
- [x] No

### Problem

#820 arms a flat 60s deadline for every window a full scan leaves unresolved. Two things let that deadline outlast the window in which a retry could still help:

- the first cold scan after launch under-resolves, because other apps' accessibility trees are still warming up;
- the app front-loads its resolution requests into its startup settling window.

One low-yield cold scan therefore bars every unresolved window past the last request the app will ever make for it, and resolution never converges for the rest of the session. `applySavedLayout` and `saveSectionOrder` skip permanently, and item moves fail with "macOS prohibits ... from being moved".

Measured on the Build DMG from run 30507663412 (`1daa5d60`, Developer ID, notarized), external non-notched main with a notched built-in secondary:

```
1 scan, 3 matches, then 101 consecutive `batch response (4 resolved)` over 3+ minutes
```

Not a signing artifact — this is a notarized build. An earlier hedge that ad-hoc identity might explain it does not hold.

### Fix

Replace the flat TTL with a per-window ladder driven by a consecutive-failure count: 5s, 15s, then the existing 60s steady state. The settling window gets three scan opportunities while a better result is still likely; a chronically unresolvable window still backs off to exactly the protection #820 added.

Failure counts reset when a window resolves and are pruned alongside `negativeUntil`, so both dictionaries stay bounded. The negative entry keeps gating scan *initiation* only — a scan started for another window still retries everything unresolved, so the late-arriving-marker behaviour #820 documents is unchanged.

The ladder is a pure function in `Shared/` so it can be pinned from `ThawTests` without instantiating the service.

Review follow-up: clear negative state when a window resolves (a scan driven by any one window can resolve others; prune unconditionally and filter the carry on windows still lacking a PID).

## PR Checklist

- [ ] I've run `swiftformat .` to keep the code style consistent.
- [x] I've run the smallest relevant test commands (list 1–2 below), e.g. `xcodebuild test …` or `swift test --package-path MenuBarModel`.
- [ ] I've documented new public APIs / non-obvious helpers.

Test commands run:

- Full ThawTests suite + archive (author-verified green)
- Ad-hoc branch build on same machine/topology: `scan 1: 3 matches → 5s retry → scan 2: 27 matches → settles 24/24`

- Independent of the divider-collapse work on #795; does not touch it. Collapse still reproduces on `1daa5d60` (`hiddenSectionHasRoom` correctly refuses the save); geometry itself remains unfixed and is tracked on #795.

## Other information

- Files: `MenuBarItemService/SourcePIDCache.swift`, `Shared/Utilities/SourcePIDNegativeCachePolicy.swift` (new), `ThawTests/SourcePIDNegativeCachePolicyTests.swift` (new)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of temporarily unresolved menu bar items with progressively adjusted retry intervals.
  * Resolved items are removed from temporary negative caching promptly, improving recovery after transient failures.

* **Tests**
  * Added coverage for retry timing, progression, boundary values, and settling-window behavior.

* **Chores**
  * Updated linting inputs to include newly covered source files.

### DIFF
--- a/MenuBarItemService/SourcePIDCache.swift
+++ b/MenuBarItemService/SourcePIDCache.swift
@@ -159,6 +159,14 @@ actor SourcePIDCache {
         /// discovered immediately.
         var negativeUntil = [CGWindowID: ContinuousClock.Instant]()
 
+        /// Consecutive full scans that have left each window unresolved.
+        /// Drives the negative-cache TTL ladder: early failures get short
+        /// deadlines so the app's startup settling window can retry while
+        /// AX trees are still warming up, repeat failures back off to the
+        /// steady-state TTL. Reset when a window resolves; pruned alongside
+        /// `negativeUntil` so it stays bounded.
+        var negativeFailures = [CGWindowID: Int]()
+
         /// Reorders the cached apps so that those that are confirmed
         /// to have an extras menu bar are first in the array.
         mutating func partitionApps() {
@@ -180,8 +188,9 @@ actor SourcePIDCache {
     /// The shared cache.
     static let shared = SourcePIDCache()
 
-    /// How long an unresolved window is barred from initiating a new scan.
-    private static let negativeCacheTTL: Duration = .seconds(60)
+    // The per-failure negative-cache deadline lives in
+    // SourcePIDNegativeCachePolicy (Shared/), so the ladder is unit-testable
+    // from ThawTests.
 
     /// Minimum interval between unresolved-diagnostic dumps for an unchanged
     /// unresolved set. The dump re-walks every app's AX tree, so repeating it
@@ -275,6 +284,7 @@ actor SourcePIDCache {
             // immediately after every application-list update.
             let now = ContinuousClock.now
             let carriedNegativeUntil = state.negativeUntil.filter { $0.value > now }
+            let carriedNegativeFailures = state.negativeFailures
 
             // Create a new state that matches the current running apps.
             state = runningApps.reduce(into: State()) { result, app in
@@ -296,7 +306,15 @@ actor SourcePIDCache {
                     }
                 }
             }
-            state.negativeUntil = carriedNegativeUntil
+            // Carry negative state only for windows that are still unresolved.
+            // A scan driven by any one window resolves every window it can, so
+            // a previously negative-cached window may now hold a PID; keeping
+            // its failure count would start its next miss partway up the
+            // ladder instead of at the first rung.
+            state.negativeUntil = carriedNegativeUntil.filter { state.pids[$0.key] == nil }
+            state.negativeFailures = carriedNegativeFailures.filter {
+                state.negativeUntil[$0.key] != nil
+            }
 
             // Log cleanup activity
             if !terminatedPids.isEmpty {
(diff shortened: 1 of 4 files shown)


diazdesandi (changes requested):
@lathe-agent-oa update template, sign the commits and address coderabbit findings.

lathe-agent-oa:
The CodeRabbit finding was right and is fixed in `66856fc0`. A scan driven by any one window resolves every window it can, so a negative-cached window can gain a PID without a scan of its own ever running. Two paths kept its failure count alive: the cleanup carry filtered on the deadline alone, and the post-scan prune sat inside a `!unresolvedWindows.isEmpty` guard that a fully-resolving scan skips. The next miss on that window then started partway up the ladder rather than at the first rung, which is the wedge this PR exists to remove. The carry now filters on the window still lacking a cached PID, and the prune runs unconditionally.

Two notes on the checks:

`metadata` is red from a stale run. It ran at 15:32:34Z against the push; I added the `Closes:` line to the body a minute later, and `pr-metadata.yml` deliberately skips `edited` events. The body satisfies the check now, but it will not re-run until the next push, and I do not have the access to re-request it. Everything else on the head commit is green.

Separately, `HookRunnerTests.timedOutHookLeavesNoDescendants()` fails on a clean `development` at `1f584161` with nothing from this branch applied — I checked it on a bare baseline before assuming otherwise. It does not surface on this PR's checks, so it is only worth knowing if you hit it elsewhere.

diazdesandi (approved):

Issue #957, opened 2026-08-18
macos-27-preview.5: duplicated/ghost menu bar icons + broken placeholder icons in Settings > Layout


kiki830621:
## Problem

Menu bar shows duplicated/ghost icons after switching to the `macos-27-preview.5` build, and Thaw's own Settings → Layout panel shows broken/placeholder icon rendering for several items.

## Environment

- macOS: 27.0 (experimental/beta, build 26A5416b)
- Thaw build: `macos-27-preview.5` (Runtime Version=27.0.0, commit `528b950`, branch `feat/macos-27-experimental`), installed manually from the GitHub release (out of the Sparkle update channel)
- Prior state: was previously on the stable Homebrew cask `1.2.0`, where Settings → Layout showed "One or more section dividers have been hidden by macOS" and prompted enabling Thaw in System Settings → Menu Bar. Switched to `macos-27-preview.5` specifically for its macOS 27 menu bar reliability fixes.

## Actual behavior

Two screenshots attached below.

1. **Menu bar overlapping/duplicated icons**: the real macOS menu bar (top strip) renders normally, but directly under/around it a large stack of overlapping, duplicated icons appears — many repeated "«" (double-chevron/collapse) placeholder icons interleaved with real app icons (cloud storage, WiFi, battery, clock, chat/AI app icons, etc.), rendered as a cascading overlap rather than a clean single row. It looks like multiple overflow/Thaw Bar popovers are stacked on top of each other rather than one clean panel.
2. **Settings → Layout ("Arrange menu bar items") panel**:
   - In the **顯示 (Visible)** section, one item renders as a hatched/diagonal-striped placeholder labeled "相機區塊" (roughly "camera block") instead of a proper icon — looks like an unresolved icon.
   - In the **隱藏 (Hidden)** section, the row is mostly filled with many repeated "«" fallback chevron icons instead of real icons, with only a handful of items showing their actual icon.

## Expected behavior

- Menu bar / Thaw Bar should render each item once, with its real icon, no duplication/overlap.
- Settings → Layout should show real icons for every managed item (or a single clear "unresolved" indicator per item), not repeated generic "«" placeholders.

## Notes

[...]

Issue #956, opened 2026-08-18
[Bug] rc4 Hidden menu items don't work when clicked on, and don't have correct labels


jamesyc:
### Pre-submission Checklist

- [x] I searched existing issues (open & closed) and found no duplicates
- [x] If this is about macOS 27 / Golden Gate, I have read
- [x] I'm using the latest app version

### What happened?

See https://github.com/thaw-app/Thaw/issues/943 for context
The behavior is much better in rc4, but still can be improved. 

The icons still load with hover text as "Menu Bar Item". The Thaw app then stutters after a few seconds, the mouse cursor jumps a few times, and the hover text gets updated... but then there are graphical glitches (notice how multiple items are highlighted). 

It's somewhat usable now, but I'd say this is "beta" quality now, up from "alpha" quality. 

### Steps to Reproduce

See https://github.com/thaw-app/Thaw/issues/943

### Expected Behavior

See https://github.com/thaw-app/Thaw/issues/943

### App Version

2.0.0-rc4

### macOS Version

26.6.x on M1 Macbook Pro

### Display Setup

Single monitor

### Logs / Console Output

[thaw_2026-08-17_17-27-27.log](https://github.com/user-attachments/files/31160224/thaw_2026-08-17_17-27-27.log)

jamesyc:
https://github.com/user-attachments/assets/7f885bee-d0ae-46b9-a42c-f6f684a05e67

diazdesandi:
Something that has caught my attention on your logs is that there seems to be an issue with MacWhisper, so:

1. Was MacWhisper actually running while this log was captured?
2. And, if it was, do you have its menu bar icon hidden or disabled in MacWhisper's own settings?

jamesyc:
No, the macwhisper process is currently dead, and I haven't started it in the past day or anything

diazdesandi:
Can you try https://github.com/thaw-app/Thaw/actions/runs/32090315006/job/95571038783 ?

jamesyc:
Still the same. 
- Thaw starts quickly in a second
- But the icons still load with hover text as "Menu Bar Item". 
- The Thaw app then stutters after a few seconds, and the hover text gets updated
- but then there are still graphical glitches (notice how multiple items are highlighted).

[thaw_2026-08-17_19-15-04.log](https://github.com/user-attachments/files/31162276/thaw_2026-08-17_19-15-04.log)

diazdesandi:
[...]

PR #821, opened 2026-07-26
fix(menubar): prevent duplicate Thaw instances from competing

alvst:
# Summary

Terminates older matching Thaw instances at launch and filters duplicate foreign control-item windows from cache handling, with focused ghost-control-window coverage.

**Scope:** One focused reliability fix so simultaneous Thaw copies do not compete over menu-bar state. Broader 2.1 feature work remains parked for a later integration branch.

> We are on the process of migrating from XCTest to Swift Test. If you are adding new tests, please use Swift Test.
> **External contributors:** before opening a PR for a bug fix or new feature, please make sure there's a corresponding issue in the [issue tracker](https://github.com/thaw-app/Thaw/issues). PRs that fix or change things that haven't been reported/agreed on may be closed without review.

Closes: N/A

## PR Type

Describe **what this change does** (not the linked issue’s request kind). Bug *reports* use `bug` on issues; bug *fixes* use `fix` on PRs.

If you tick **Feature** or **Refactor** and touch more than ~20 files, please mention why this can’t be split.

- [x] Bug fix
- [ ] CI/CD
- [ ] Documentation
- [ ] Feature
- [ ] Enhancement
- [ ] Performance improvement
- [ ] Refactor
- [x] Test addition or update
- [ ] Other (please describe)

## Area

**Product surfaces** (optional when the change is not about the app UI). Path-based labeling also applies.

- Use **PR Type** for *what* changed (`CI/CD`, `Documentation`, `Other` / chore, etc.).
- Use **`ops`** for *where* when it is repo operations: CI, release, GitHub hygiene, scripts, lint/sonar config — not a product surface.

- [x] menubar
- [ ] icebar
- [ ] layout
- [ ] appearance
- [ ] settings
- [ ] onboarding
- [ ] permissions
- [ ] profiles
- [ ] hotkeys
- [ ] updates
- [ ] ops

## Does this PR introduce a breaking change?

- [ ] Yes - if yes, please describe the impact and migration path
- [x] No

## What is the new behavior?

At launch, older matching Thaw instances are terminated so only one copy manages the menu bar. Duplicate foreign control-item windows are filtered from cache handling. Adds focused ghost-control-window test coverage (Swift Testing).

## PR Checklist

- [x] I've built and run the app locally and verified that it works as expected.
- [ ] I've run `swiftformat .` to keep the code style consistent.
- [x] I've run the smallest relevant test commands (list 1–2 below), e.g. `xcodebuild test …` or `swift test --package-path MenuBarModel`.
[...]

PR #962, opened 2026-08-19
fix(layout): stop a degenerate window from discarding hidden-item preview batches

### LINKED ISSUE #963: [Bug] Settings → Layout shows no icons for hidden items after an ungraceful quit (orphaned Control Center slot)
### Pre-submission Checklist

- [x] I searched existing issues (open & closed) and found no duplicates
- [x] If this is about macOS 27 / Golden Gate, I have read
- [x] I'm using the latest app version

### What happened?

In Settings → Layout, the Hidden and Always-Hidden rows render completely empty — no icons, no placeholders — while Visible renders normally. The items are fine (enumerated, managed, hiding/revealing works); only the previews are missing, which makes the layout editor unusable for arranging anything not already visible in the menu bar.
Image

# Cause
A single zero-area window in a capture batch discards the whole batch.

refreshImages captures the batch as one composite image and slices it per item using each window's offset within the batch's bounds union, guarding first that the composite's width equals that union scaled. A zero-width/height window contributes no pixels — the capture API omits it — but its origin still widens the union. Parked off-screen it drags the union across the whole gap, the widths can't match, and everything in the batch is thrown away.

In my case the zero-area windows were orphaned Control Center hosting slots. Thaw's status items are hosted by Control Center, so those windows are owned by the Control Center process, not Thaw:

wid=10980  Control Center  x=1503   w=33    h=33  name=Thaw.ControlItem.Visible
wid=10981  Control Center  x=-3513  w=5016  h=33  name=Thaw.ControlItem.Hidden
wid=10982  Control Center  x=-8789  w=5016  h=33  name=Thaw.ControlItem.AlwaysHidden
wid=10267  Control Center  x=-8789  w=0     h=0   name=com.stonerl.Thaw   <-- orphan
wid=10268  Control Center  x=-8789  w=0     h=0   name=com.stonerl.Thaw   <-- orphan

The last two belong to an earlier Thaw process — their window IDs are lower than the running instance's control items (10980–10982) and its Thaw Bar (10975), and they appear in logs from before that instance started. Control Center owns them, so they outlive Thaw restarts. They're named with the bundle ID rather than Thaw.ControlItem.*, so ControlItemPair doesn't recognise them as control items and doesn't strip them — and they ride into the hidden/always-hidden batch.

### Steps to Reproduce

Needs an orphaned slot to exist, so it can't be triggered from the UI alone:

1. Force-quit Thaw ungracefully a few times (pkill -f Thaw, a crash, a force quit) — each can leave a hosting slot behind.
2. Relaunch and hide some items.
3. Open Settings → Layout.

### Expected Behavior

Expected: Hidden/Always-Hidden rows show each item's icon.
Actual: both rows are empty.

### App Version

2.0.0-rc.4

### macOS Version

26.5.2

### Display Setup

Single monitor

### Logs / Console Output

[thaw_2026-08-18_21-05-02.log](https://github.com/user-attachments/files/31207129/thaw_2026-08-18_21-05-02.log)

### Additional Context

Fix:  #962

alvst:
**Base:** `86fd78b6` (post-2.0.0-rc.4)

Closes: #963

## Symptom

In **Settings → Layout**, the Hidden and Always-Hidden rows render completely
empty — no icons, no placeholders — while the Visible row renders normally. The
items themselves are fine: they're enumerated, managed, and moving between
sections works. Only the previews are missing.

It happens whenever a hidden-section capture batch contains a **zero-area
window**. In the case investigated here those were orphaned Control Center
hosting slots left behind by an earlier Thaw process — see *How the state
arises*. The state is transient: once Control Center reclaims the slots, stock
`development` renders previews normally again.

Changing the **Section divider style** makes no difference (verified with `None`
and `Chevron`, identical failure) — that was an early wrong guess on my part.

## Cause

`refreshImages` captures a batch of item windows as a single composite image,
then slices it per item using each window's offset within the batch's bounds
union:

```swift
boundsUnion = boundsUnion.union(bounds)          // built over every item
…
let expectedWidth = boundsUnion.width * scale
guard CGFloat(compositeImage.width) == expectedWidth else { … skipping }
…
x: (bounds.origin.x - boundsUnion.origin.x) * scale
```

A window with zero width or height breaks that arrangement. The capture APIs
leave it out of the composite, so it contributes no pixels — but its origin
still widens the union. Parked off-screen, it drags the union across the entire
gap, so the composite's width can no longer match, and the batch is discarded.

### How the state arises

Thaw's status items are hosted by Control Center, so all of its control item
windows are owned by the Control Center process rather than by Thaw. A direct
window-list dump during the failure:

```
wid=10980  Control Center  x=1503   w=33    h=33  name=Thaw.ControlItem.Visible
wid=10981  Control Center  x=-3513  w=5016  h=33  name=Thaw.ControlItem.Hidden
wid=10982  Control Center  x=-8789  w=5016  h=33  name=Thaw.ControlItem.AlwaysHidden
wid=10267  Control Center  x=-8789  w=0     h=0   name=com.stonerl.Thaw   <-- orphan
wid=10268  Control Center  x=-8789  w=0     h=0   name=com.stonerl.Thaw   <-- orphan
```

The last two are hosting slots orphaned by an **earlier** Thaw process: their
window IDs are lower than the running instance's control items (10980-10982) and
lower than its Thaw Bar (10975), and they appear in logs from before that
instance started. Because Control Center owns them, they outlive Thaw restarts.
Their name is the bundle ID rather than `Thaw.ControlItem.*`, so
`ControlItemPair` does not recognise them as control items and does not strip
them — and they ride into the hidden/always-hidden batch.

I produced them by force-quitting Thaw repeatedly while debugging something
else; a crash or an ungraceful exit does the same thing to a user. They are not
permanent: `killall ControlCenter` reclaims them.

### Verified in all three states

| Orphan slots | Build | Result |
| --- | --- | --- |
| present | `development` | every batch discarded, rows empty (32 discards, 0 refreshes) |
| reclaimed (after `killall ControlCenter`) | `development` | previews render, `refreshImages: ✓ updated 4/4 items` |
| present | with this patch | previews render |

So this PR does not change behaviour in the healthy state; it makes the capture
geometry immune to a degenerate window, whatever produced it.

## Measurement

Instrumented `refreshImages` on an affected machine (hidden=8, always-hidden=2,
2× notched display):

```
composite=520x66  scale=2.0  unionX=-3774  unionW=5276  sumW=260
items= com.apple.controlcenter:Item-0@x=1242,w=42
     | com.apple.controlcenter:Item-0:1@x=1284,w=33
     | com.apple.controlcenter:Item-0:2@x=1317,w=34
     | com.apple.controlcenter:NowPlaying@x=1351,w=33
     | com.apple.controlcenter:Sound@x=1384,w=38
     | com.apple.controlcenter:FocusModes@x=1422,w=38
     | com.apple.controlcenter:Battery@x=1460,w=42
     | com.apple.controlcenter:com.stonerl.Thaw@x=-3774,w=0
     | com.apple.controlcenter:com.stonerl.Thaw:1@x=-3774,w=0
```

The seven real items are contiguous — x=1242→1502, 260pt — and the capture
returns exactly those at 520px (260 × 2). The two zero-width dividers at
x=−3774 inflate the union to 5276pt, so the guard compares **520 against
10552** and skips.

Across a session of diagnostic logs: **32 batches discarded, zero successful
hidden-section refreshes.** Deterministic, not a race — the guard cannot pass
while a degenerate window is in the batch.

## Fix

Filter degenerate bounds out of the batch before they reach the union. Such a
window could never have produced a preview anyway: `cropping(to:)` on an empty
rect returns nil, so it was always destined to be skipped a few lines later —
just after it had already corrupted the geometry for every other item.

`isCapturableBounds` is a pure predicate so the rule is testable without a live
window server. 5 tests, including one that reconstructs the geometry above and
pins both numbers (5276pt unfiltered vs 260pt filtered).

**On placement:** the filter sits where the union is built, not at the call site
that assembles the batch. Excluding collapsed dividers earlier would fix this
particular reproduction, but *any* zero-area window breaks composite slicing the
same way regardless of how it got there, so the invariant belongs next to the
geometry that depends on it. Happy to move it if you'd rather the batch never
contain them in the first place.

## Testing

- Full suite: **2,410 tests in 306 suites, 0 failures** (`xcodebuild test`,
  Debug, macOS).
- 5 new tests; suite count +5 from the current `development` baseline.
- Reproduced and instrumented on the affected machine; the measurement above is
  from that build. Before/after screenshots to follow — this is why the PR is
  open as a draft.

## Environment

- macOS 26.5.2 (25F84), Apple M1 Max, built-in Liquid Retina XDR (2× scale, notched)
- hidden = 8 items, always-hidden = 2
- Independent of Section divider style (`None` and `Chevron` both fail)

## Diffstat

```
 .gitignore                                                      |  1 +
 Thaw/MenuBar/MenuBarItems/MenuBarItemImageCache.swift           | 27 +++++++
 ThawTests/MenuBar/Items/CapturableBoundsTests.swift             | 77 ++++++++++++++++
 3 files changed, 105 insertions(+)
```

(The `.gitignore` line is `default.profraw`, a coverage artifact the test run
drops in the repo root.)

## Possibly related

- **#956** — "rc4 Hidden menu items don't work when clicked on, and don't have
  correct labels" mentions hidden items showing "Menu Bar Item" instead of real
  labels. If those rows are also missing their previews, this may be part of it,
  though the click and cursor-jump symptoms in that report are separate.
- **#957** — closed as a duplicate of #687; that one is on
  `macos-27-preview.5` and shows *placeholder* chevrons rather than empty rows,
  so it looks like a different failure.

## Follow-up worth considering separately

The orphaned slots slip past the ghost-control-item filter added in #821,
because that keys on the `Thaw.ControlItem.*` naming while these carry the
bundle ID. Dropping Control-Center-owned zero-size windows titled with our own
bundle ID at enumeration would remove them before anything downstream — the
image cache, the layout editor, the item cache — ever sees them. That seems
like the better root fix, and complementary to this one rather than a
replacement: composite slicing should not be fragile to a zero-area window
regardless of how one got into the list. Happy to open an issue for it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved menu bar image capture by ignoring collapsed or empty menu items.
  * Prevented invisible dividers from distorting capture dimensions and causing valid image batches to be discarded.

* **Tests**
  * Added coverage for zero-sized bounds, empty rectangles, and collapsed off-screen dividers.

* **Chores**
  * Excluded profiling output files from version control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ plans/
 
 # Local badge mockups
 .tmp-badge-preview/
+default.profraw
--- a/Thaw/MenuBar/MenuBarItems/MenuBarItemImageCache.swift
+++ b/Thaw/MenuBar/MenuBarItems/MenuBarItemImageCache.swift
@@ -236,6 +236,15 @@ final class MenuBarItemImageCache: @unchecked Sendable {
     /// Minimum spacing enforced between offscreen SkyLight batch captures.
     private static let minSkyLightBatchInterval: Duration = .seconds(1)
 
+    /// Whether a window's bounds can contribute to a composite capture.
+    ///
+    /// Both dimensions must be positive: the capture APIs omit degenerate
+    /// windows from the composite, so including one only corrupts the
+    /// geometry the composite is sliced with.
+    static nonisolated func isCapturableBounds(_ bounds: CGRect) -> Bool {
+        bounds.width > 0 && bounds.height > 0
+    }
+
     /// Timestamp of the last visible-section SCK capture, used to rate-limit
     /// the on-screen path the same way the offscreen one already is.
     private var lastSCKRefreshAt: ContinuousClock.Instant?
@@ -1262,6 +1271,25 @@ final class MenuBarItemImageCache: @unchecked Sendable {
 
         for item in items {
             guard let bounds = Bridging.getWindowBounds(for: item.windowID) else {
+                continue
+            }
+            // Degenerate windows must not reach the union. A zero-width or
+            // zero-height window contributes nothing to the composite — the
+            // capture APIs drop it, and `cropping(to:)` on an empty rect
+            // returns nil — but including it still corrupts the geometry the
+            // composite gets sliced against: parked off-screen it drags
+            // `boundsUnion` across the whole gap to its position, so the
+            // expected-width check below compares the composite of the real
+            // items against a union thousands of points wide and discards
+            // every batch.
+            //
+            // This holds whatever produced the degenerate bounds; the union
+            // and the slice loop below read the same storage, so a window that
+            // cannot contribute geometry must enter neither.
+            guard Self.isCapturableBounds(bounds) else {
+                MenuBarItemImageCache.diagLog.debug(
+                    "refreshImages: skipping degenerate bounds for \(item.logString) (\(bounds.width)x\(bounds.height))"
+                )
                 continue
             }
             windowIDs.append(item.windowID)
--- a/ThawTests/MenuBar/Items/CapturableBoundsTests.swift
+++ b/ThawTests/MenuBar/Items/CapturableBoundsTests.swift
@@ -1,0 +1,79 @@
+//
+//  CapturableBoundsTests.swift
+//  Project: Thaw
+//
+//  Copyright (Ice) © 2023–2025 Jordan Baird
+//  Copyright (Thaw) © 2026 Toni Förster
+//  Licensed under the GNU GPLv3
+
+import CoreGraphics
+import Testing
+@testable import Thaw
+
+/// Characterizes which window bounds may take part in a composite capture.
+///
+/// `refreshImages` captures a batch of item windows as one image and slices it
+/// per item using each window's offset within the batch's bounds union. A
+/// degenerate window — zero width or zero height — breaks that: the capture
+/// APIs drop it from the composite, so it adds no pixels, while including it in
+/// the union corrupts the geometry the composite gets sliced against. Parked
+/// off-screen it widens the union across the whole gap, the composite's width
+/// no longer matches, and the batch is discarded. That is the difference
+/// between previews rendering and the Hidden rows appearing empty.
+///
+/// Which item produced the degenerate bounds does not matter to any of this.
+@Suite("Capturable bounds")
+struct CapturableBoundsTests {
+    @Test("An ordinary menu bar item window is capturable")
+    func ordinaryWindowIsCapturable() {
+        #expect(MenuBarItemImageCache.isCapturableBounds(CGRect(x: 1242, y: 0, width: 42, height: 33)))
+    }
+
+    @Test("A zero-width window is not capturable")
+    func zeroWidthIsNotCapturable() {
+        // A degenerate window parked off-screen, as observed in the field.
+        #expect(!MenuBarItemImageCache.isCapturableBounds(CGRect(x: -3774, y: 0, width: 0, height: 33)))
+    }
+
+    @Test("A zero-height window is not capturable")
+    func zeroHeightIsNotCapturable() {
+        #expect(!MenuBarItemImageCache.isCapturableBounds(CGRect(x: 1242, y: 0, width: 42, height: 0)))
+    }
+
+    @Test("A fully empty rect is not capturable")
+    func emptyRectIsNotCapturable() {
+        #expect(!MenuBarItemImageCache.isCapturableBounds(.zero))
+        #expect(!MenuBarItemImageCache.isCapturableBounds(.null))
+    }
+
+    /// The regression itself, in the geometry the field log recorded: seven
+    /// contiguous on-screen items plus two degenerate windows at x=-3774. With
+    /// those filtered, the union matches the composite the capture returns
+    /// (260pt → 520px at 2x); without them it is 5276pt wide and every batch is
+    /// thrown away.
+    @Test("Degenerate windows no longer widen the batch union")
+    func degenerateWindowsDoNotWidenTheUnion() {
+        let real = [
+            CGRect(x: 1242, y: 0, width: 42, height: 33),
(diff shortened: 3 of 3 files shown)


diazdesandi (changes requested):
Keep the guard where you put it. Filtering at the union site is the right call: the batch and the slice loop read the same storage, so a window that cannot contribute geometry should never enter either. Moving it to the call site would leave the two halves free to disagree later.

One thing to change, and it is labels only. The PR body retracts the collapsed-divider explanation, but the retracted version still ships in four places:

1. PR title
2. The comment above the new guard in MenuBarItemImageCache.swift
3. CapturableBoundsTests.swift: the suite doc comment, the test name collapsedDividersDoNotWidenTheUnion, and that test's own doc comment
4. The // collapsed divider parked off-screen, as observed in the field comment in the fixture

Anyone reading this in six months gets the diagnosis you already withdrew. Suggested framing for the replacement: a zero-width or zero-height window is degenerate, the capture APIs drop it from the composite, and including it in the union corrupts the geometry the composite gets sliced against. That holds regardless of which item produced the degenerate bounds, which is why it is the better description.

The measurements and the geometry are correct as written. Nothing in the code needs to move.

alvst:
Updated the title. The title was my originally-suspected cause but through more digging, I found the actual cause/fix and forget to update the title. 

Also pushed the same reframing everywhere else it shipped: the guard comment, the suite doc comment, the test (collapsedDividersDoNotWidenTheUnion → degenerateWindowsDoNotWidenTheUnion) and its doc, the fixture comment, and the commit message. They all describe the invariant now — a zero-width or zero-height window is dropped from the composite by the capture APIs, and including it in the union corrupts the geometry the composite gets sliced against, regardless of which item produced the bounds.

diazdesandi (approved):

PR #735, opened 2026-06-21
feat(menubar): add menu bar item triggers

alvst:
# Summary

Adds trigger-based menu bar item visibility so items can appear only when specific system, app, power, or network conditions are met, including the behavior requested in #422.

**Scope:** Menu bar item triggers feature (settings + runtime). Large surface; still integrating toward sanitize-dev.

> **External contributors:** before opening a PR for a bug fix or new feature, please make sure there's a corresponding issue in the [issue tracker](https://github.com/stonerl/Thaw/issues). PRs that fix or change things that haven't been reported/agreed on may be closed without review.

Closes: #422

## PR Type

Describe **what this change does** (not the linked issue’s request kind). Bug *reports* use `bug` on issues; bug *fixes* use `fix` on PRs.

If you tick **Feature** or **Refactor** and touch more than ~20 files, please mention why this can’t be split.

- [ ] Bug fix
- [ ] CI/CD
- [x] Documentation
- [x] Feature
- [ ] Enhancement
- [ ] Performance improvement
- [ ] Refactor
- [x] Test addition or update
- [ ] Other (please describe)

## Area

**Product surfaces** (optional when the change is not about the app UI). Path-based labeling also applies.

- Use **PR Type** for *what* changed (`CI/CD`, `Documentation`, `Other` / chore, etc.).
- Use **`ops`** for *where* when it is repo operations: CI, release, GitHub hygiene, scripts, lint/sonar config — not a product surface.

- [x] menubar
- [ ] icebar
- [ ] layout
- [ ] appearance
- [x] settings
- [ ] onboarding
- [ ] permissions
- [ ] profiles
- [x] hotkeys
- [ ] updates
- [ ] ops

## Does this PR introduce a breaking change?

- [ ] Yes - if yes, please describe the impact and migration path
- [x] No

## What is the new behavior?

Adds trigger infrastructure, Triggers settings, and battery/power-based conditions. Developer settings gate additional sources. Hardens trigger-driven cursor/menu-bar movement so stale checks do not keep posting synthetic movement after a condition is invalid.

## PR Checklist

- [x] I've built and run the app locally and verified that it works as expected.
- [x] I've run `swiftformat .` to keep the code style consistent.
- [x] I've run the smallest relevant test commands (list 1–2 below), e.g. `xcodebuild test …` or `swift test --package-path MenuBarModel`.
- [x] I've added tests for new behavior (if applicable).
- [ ] I've documented new public APIs / non-obvious helpers.
- [x] I've updated documentation as needed.
[...]

PR #965, opened 2026-08-22
feat(menubar): menu bar item triggers (updated port of #735)

### LINKED ISSUE #422: [Feature Request]:add triggers

### Description

Add triggers, such as displaying the icon only when the Wi-Fi connection is lost.

### Additional Information

_No response_

alvst:
## Summary

Reopens the menu bar item triggers feature from #735, ported onto current `development` and substantially extended. Triggers conditionally reveal or hide a menu bar item — e.g. *show Battery in the Visible bar while it's below 20%, otherwise keep it in the Hidden bar*.

All 27 condition sources from #735 are included (battery/power, frontmost app, network/VPN/SSID, Bluetooth, audio output, displays, schedules, Focus, geofence, camera/mic, script result, icon-change detection), feature-flag gated per source via a new **Dev Mode Flags** pane so they can be enabled and stabilized one at a time. Battery/power conditions are always available.

## What's new beyond the original port

- **Energy Mode conditions** — Low Power / Not Low Power / Automatic / High Power. High Power has no public API; it's read from `State:/IOKit/PowerManagement/CurrentSettings` via a new `EnergyModeMonitor` (SCDynamicStore-observed, since it changes without a notification) and offered only on Macs that publish the key. Legacy `.lowPowerMode` conditions keep decoding.
- **"None of" matching** — conditions can combine as All of / Any of / None of, with None of usable on a single condition for plain negation. Summaries are phrased so a joined list can't read as its opposite.
- **Bluetooth device picker** — replaces the free-text device name with a picker over the paired-device names the matcher actually compares against. A device's classic Bluetooth name can differ from what System Settings shows (AirPods Pro report a generic model name), so a typed name could silently never match. Enumeration runs off the main actor (IOBluetooth blocks and can wait behind a TCC prompt), and `NSBluetoothAlwaysUsageDescription` is added — its absence crashed any unauthorized Bluetooth access, including the original port's condition sampler.
- **Layout-editor ownership** — a trigger-owned item draws dimmed with a bolt badge, its tooltip names the owning trigger, and the trigger row explains that the saved layout no longer applies to it. Ownership updates live (stored `@Observable` set), resolves exact instances so same-title siblings are distinguished, and the badge, tooltip, and priority plan answer from one predicate.
- **Hardening** — settle debounce keeps its deadline instead of restarting on every state publication (a jittery source could previously postpone an apply indefinitely); reveal notifications post only after a physical move; network state is seeded synchronously so an offline launch can't briefly run the "connected" action; `PowerSourceMonitor` no longer mixes fields across battery and UPS sources; mic-in-use reads input streams rather than the device-wide flag that false-positived on duplex hardware; saved-layout applies exclude trigger-owned items.
- Removed the original port's carve-out that silently kept the Control Center Battery item from ever hiding — its premise (that concealing Battery flips the system's own *Show in Menu Bar* setting) did not reproduce; a test now pins the corrected behavior.

## Verification

`ThawTests` passes in full (2406 tests / 305 suites), including ~200 trigger-specific tests covering condition evaluation, priority planning, ownership resolution, combinator semantics, legacy decoding, and the script-runner process-group contract. SwiftFormat/SwiftLint clean apart from violations that pre-date the branch.

## Additional Considerations: triggers vs. saved layout / profiles

While a trigger is enabled, its target item is currently frozen out of layout customization: it's shielded from `saveSectionOrder` persistence, so the user can't record a new home for it in any profile, and per-profile arrangement of that item is impossible without disabling the trigger first. The shields also have to be re-applied at every layout write site (`applySavedLayout` has the exclusion; `applyProfileLayout` currently doesn't, so a Focus-driven profile switch can briefly fight a profile-dependent trigger).

We think the right fix is a structural one: **separate saved layout (user intent — per-profile, always editable, what profiles capture) from trigger placement (a runtime overlay that never writes to intent and restores it on release).** The release path already restores intent, and persistence already refuses trigger positions — the semantics are effectively there; only the representation conflates the two, which is why each write site needs its own shield. With the split, profiles stay fully customizable while triggers run, per-profile ordering works, and the write-site exclusion class disappears. Cost: the Layout pane would show intent rather than the live bar for trigger-held items (badge explaining the difference), and it touches the layout-apply core, so we'd want to design it up front rather than patch it in.

Alternatives considered: per-profile section overrides on triggers (couples the two data models, adds a config matrix, and still leaves per-profile ordering frozen — and compound conditions like *profile is Work AND battery below 20%* already express per-profile behavior), or leaving the toggle-off-then-arrange workflow documented as-is.

Curious your thoughts on this and if it should be handled as part of this PR or a separate one. 

## Additional Notes: 
Due to the number of interrelated features, this is hard to fit into the line/file guideline (20 files, 500 lines) — the model, monitors, manager, and UI all reference each other, so slices wouldn't stand alone. It is substantially smaller and more focused than #735: everything here is the trigger feature, each source is gated behind its own feature flag, and there are ~200 trigger-specific tests.

Closes: #422

## Summary by CodeRabbit

* **New Features**
  * Added conditional triggers that reveal, hide, and reposition menu bar items based on battery, apps, network, Focus, location, Bluetooth, scripts, images, and other system conditions.
  * Added Triggers and Developer settings panes with condition editors, priorities, diagnostics, and experimental source controls.
  * Added trigger status indicators, ownership tooltips, notifications, Energy Mode and thermal conditions, and emergency controls to disable all trigger features.
* **Improvements**
  * Saved layouts now better preserve manually arranged items affected by triggers.
  * Added permission guidance for supported system data sources and searchable settings entries.


diazdesandi (approved):

diazdesandi:
Approved, going to merge to start testing it together with the new XPC and get it ready to ship

Issue #732, opened 2026-06-19
[Feature Request] Split logs by size/time


nk-tedo-001:
### Pre-submission Checklist

- [x] I searched existing issues (open & closed) and found no duplicates

### Problem / Use Case

I can't upload the big file because of the GitHub validation. Compressing it is an additional action.

> File size too big: 25 MB are allowed, 26 MB were attempted to upload.

### Proposed Solution

For example, split the files by 10 MB or each hour, but no more than two days.

diazdesandi:
This limitation comes from GitHub’s own attachment size cap, and Thaw doesn’t control how GitHub uploads or stores files. Implementing automatic log splitting on the Thaw side wouldn’t change GitHub’s validation behavior for already-generated artifacts, so we can’t reliably fix this from within the action itself. 

Since this isn’t something we can address in Thaw directly, I’ll close this issue.

Makishima:
@diazdesandi I know it's a GitHub limitation, but it could easily be fixed by the Thaw side just by rotating the logs by size as many apps can do.

diazdesandi:
@Makishima I get where you’re coming from, but our focus right now is on that core experience and the new macOS 27 implementation. We’re a very small team with an existing backlog of open bugs and feature requests that directly affect day‑to‑day use, so we have to reserve our bandwidth for those.

Given that, adding per‑run log rotation logic inside this action isn’t something we can realistically focus on. For now we’re going to keep this as “not planned," but if someone from the community wants to propose an implementation via PR, we’re happy to review it.

Issue #899, opened 2026-08-06
[Bug] RC2.1 destroys my cursor moves


warmup72:
### Pre-submission Checklist

- [x] I searched existing issues (open & closed) and found no duplicates
- [x] If this is about macOS 27 / Golden Gate, I have read
- [x] I'm using the latest app version

### What happened?

After switching from rc 2.0 und rc 2.1 the space between the icons is set to Standard and I have no chance to reduce this back.
But on my macOS 26.6 I can´t use the new release because the cursor is moving unmotivated, went big and smal sometimes I see 3 ones - activating the log was a challenge:-)

I´m back on rc 2.0 :-(

### Steps to Reproduce

brew upgrade
Start thaw and see the problem

### Expected Behavior

should be good

### App Version

rc 2.1

### macOS Version

26.6

### Display Setup

Single monitor

### Logs / Console Output

[thaw_2026-08-06_14-03-56.log](https://github.com/user-attachments/files/30792643/thaw_2026-08-06_14-03-56.log)

diazdesandi:
Thanks for the report, I am looking into it, it's likely a duplicate of #895 , closing it until it's clear

warmup72:
Thank you very much - I'm not sure about the duplicate - for me the description of 895 is very complicated - I can't follow that - but if you think and maybe my log will help, then I'll try again when 895 is fixed :-)

Thank you for the great work!

diazdesandi:
Can you try https://github.com/thaw-app/Thaw/actions/runs/31127665694/artifacts/8974810339 ?

warmup72:
That was now the killer attempt :-(

After starting the version I can no longer operate my MBP - mouse pointer is no longer controllable.

But the bad thing - even a switch to rc 2.0 is no longer possible.

Can you tell me what I need to delete so that at least sic 2 works again.

warmup72:
Maybe that helps - I have now after the rc2.0 spun the same and additionally also the spaces between the icons had set to max. it with the twitching pointer to reduce this space and clicked on "Broadcast to all screens".

That helped. Problem is fixed again for rc 2.0. In my opinion, it seems to have something to do with this distribution of the icons - Could it be that because of this large free space with 20-30 icons something overflows in the menu line?

diazdesandi:
There were issues with the spacing and this other issue before. Can you try this one? https://github.com/thaw-app/Thaw/actions/runs/31161594220

warmup72:
[...]

PR #738, opened 2026-06-22
feat(logging): rotate diagnostic logs by size and time

nk-tedo-001:
# Summary

Adds automatic rotation of `~/Library/Logs/Thaw` diagnostic logs so a single file stays under GitHub’s 25 MB upload limit. Rotation by size (default 10 MB) and optional time interval; age-based retention.

**Scope:** Diagnostic log rotation + Advanced settings + profile/reset plumbing; main app owns rotation and re-points XPC.

> **External contributors:** before opening a PR for a bug fix or new feature, please make sure there's a corresponding issue in the [issue tracker](https://github.com/stonerl/Thaw/issues). PRs that fix or change things that haven't been reported/agreed on may be closed without review.

Closes: #732

## PR Type

Describe **what this change does** (not the linked issue’s request kind). Bug *reports* use `bug` on issues; bug *fixes* use `fix` on PRs.

If you tick **Feature** or **Refactor** and touch more than ~20 files, please mention why this can’t be split.

- [ ] Bug fix
- [ ] CI/CD
- [ ] Documentation
- [x] Feature
- [ ] Enhancement
- [ ] Performance improvement
- [x] Refactor
- [x] Test addition or update
- [ ] Other (please describe)

## Area

**Product surfaces** (optional when the change is not about the app UI). Path-based labeling also applies.

- Use **PR Type** for *what* changed (`CI/CD`, `Documentation`, `Other` / chore, etc.).
- Use **`ops`** for *where* when it is repo operations: CI, release, GitHub hygiene, scripts, lint/sonar config — not a product surface.

- [ ] menubar
- [ ] icebar
- [ ] layout
- [ ] appearance
- [x] settings
- [ ] onboarding
- [ ] permissions
- [x] profiles
- [ ] hotkeys
- [ ] updates
- [x] ops

## Does this PR introduce a breaking change?

- [ ] Yes - if yes, please describe the impact and migration path
- [x] No

## What is the new behavior?

- Size rotation (default 10 MB) + optional time rotation (off/hourly/daily).
- Age-based retention (default 2 days) + count cap.
- `configureLogging(filePath:)` optional — `nil` disables logging.
- Advanced → Diagnostics settings; carried in profiles and reset.
- Unit tests for retention/filename cores.

## PR Checklist

- [x] I've built and run the app locally and verified that it works as expected.
- [x] I've run `swiftformat .` to keep the code style consistent.
[...]

PR #974, opened 2026-08-25
feat(logging): rotate diagnostic logs by size and time

nk-tedo-001:
# Summary

Diagnostic logging writes one file per session, and that file grows until the session ends. A long session leaves a log that is too large to attach to a bug report, which is what #732 asks about.

This adds rotation. A new file starts once the current one reaches a size limit, and optionally on a schedule. Retention now keeps files by age instead of keeping a fixed five.

This is the work from #738, rebuilt against `release/v2.1.0` as requested there. The old branch no longer applied: the base moved on by 227 commits, the tests moved to Swift Testing, and the XPC protocol lost its single-window `sourcePID` request.

## Linked issue (required)

Closes: #732

## PR Type

- [x] Feature

## Area

- [x] settings

- [x] No

The app and the bundled XPC service ship together, so the `configureLogging` payload change cannot meet an older peer. Settings read back with `decodeIfPresent` and their defaults, so existing preferences load unchanged.

## What is the new behavior?

Rotation starts a new log file once the current one reaches 10 MB, and can also rotate hourly or daily. Retention deletes files older than two days and never keeps more than 50, replacing the previous "keep the five newest". All three are settings, in Advanced under Diagnostics.

The app owns rotation, because only the process that mints a file may mint the next one. Both processes still prune the directory they share, so the retention policy now travels to `MenuBarItemService` along with the log path. That path became optional: `nil` turns the service's file logging off, which also fixes a smaller bug where toggling diagnostics after launch never reached the service.

One sender in `MenuBarItemService.Connection` handles every reason the service needs telling: a rotation, a settings change, a failed push, or a dropped session. It reads the current path when it sends and keeps a single request in flight, so a late notification cannot leave the two processes on different files. A failed push is retried, and the service answers with an error when it cannot open the file rather than reporting success it did not achieve.

A few things came out of review while porting:

Log files are now created with `0o600` and `O_NOFOLLOW`. They carry window titles and process names, so world-readable was the wrong default.

Rotation opens the new file before closing the old one. If the open fails, the current segment keeps taking writes instead of logging stopping.

A policy that arrives over XPC is clamped before use. A build signed without a team identifier accepts messages from any local process, and an unchecked policy there could delete every log (negative retention puts the cutoff in the future) or trap the arithmetic that decides what to keep (`Int.min`).

Scheduled rotation measures elapsed time with `ContinuousClock`, so changing the system clock does not skip or delay it.

## PR Checklist

- [ ] This PR targets the `development` branch.
- [x] If this PR changes dependencies / lockfiles, `dependency-sca` is green.

The branch targets `release/v2.1.0`, as asked in #738.

Test commands run:

- `xcodebuild test -project Thaw.xcodeproj -scheme Thaw -configuration Debug -destination 'platform=macOS'` (2547 tests in 324 suites, all passing)
- `xcodebuild test ... -only-testing:ThawTests/DiagnosticLoggerFileTests -only-testing:ThawTests/DiagnosticLoggerRetentionTests`

Five existing tests in `DiagnosticLoggerFileTests` changed, because the behavior they pin changed on purpose:

The four rotation tests pinned "prune down to the five newest". They now pin the age based window, using the same seeded files, and one of them still covers the case where an undeletable file must not stop the rest of the prune.

"Re-attaching to the same file appends a second header" is now two tests. Attaching to a file that already holds lines still appends rather than truncating, but re-attaching to the file already open does nothing. The app re-sends the current path on rotation and on retry, and reopening it would stamp a second header and a stop footer into the file being written.

Deliberately out of scope:

The new settings are not carried in profiles. #899 removed `enableDiagnosticLogging` from the snapshot because a profile switch turned logging off in the middle of the capture it was meant to record, and these settings have the same problem.

New UI strings are not added to `Localizable.xcstrings`. Running the extraction here rewrites the whole catalog, which seemed worse than leaving three strings for your own pass.

The path check in `Listener` is still open to TOCTOU: it resolves and prefix-checks the path before the open, and `O_NOFOLLOW` only covers the last component. Closing that properly means restricting the request to a basename and opening with `openat` against a directory descriptor. It is the existing check rather than something this PR introduces, so it seemed like a separate change.

## Other information

The controls sit in Advanced under Diagnostics, next to the toggle they belong to, since that is where the section lives on this branch. If they belong in the Tools tab from your screenshot in #738, they should move with the section rather than separately.

Verified by build and the full test suite. The settings layout was also checked in a local run of an earlier revision of this work.

---
<a href="https://app.blacksmith.sh/thaw-app/codesmith/Thaw/pr/974"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790235428&installation_model_id=431242&pr_number=974&repository=thaw-app%2FThaw&return_to=https%3A%2F%2Fgithub.com%2Fthaw-app%2FThaw%2Fpull%2F974&signature=e73e3d25dc723a2d8f774cd8d629e50a1502d781d9f8675bcda8124946904d81"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/MenuBarItemService/Listener.swift
+++ b/MenuBarItemService/Listener.swift
@@ -42,7 +42,19 @@ final nonisolated class Listener: @unchecked Sendable {
             case .start:
                 diagLog.debug("Listener received start request")
                 return .start
-            case let .configureLogging(filePath):
+            case let .configureLogging(filePath, rotationPolicy):
+                if let rotationPolicy {
+                    // Prune by the app's retention settings, not this target's
+                    // defaults: both processes share one log directory.
+                    DiagnosticLogger.shared.setRotationPolicy(rotationPolicy)
+                }
+                guard let filePath else {
+                    // The app turned file logging off; stop writing to the
+                    // shared file rather than holding it open.
+                    DiagnosticLogger.shared.isEnabled = false
+                    diagLog.debug("Listener disabled diagnostic logging")
+                    return .configureLogging
+                }
                 // Only attach to files inside the app's approved log
                 // directory. The path arrives from the XPC peer, and in
                 // teamless (ad-hoc) builds the listener has no peer
@@ -58,7 +70,14 @@ final nonisolated class Listener: @unchecked Sendable {
                     )
                     return nil
                 }
-                DiagnosticLogger.shared.attachToFile(at: requested)
+                guard DiagnosticLogger.shared.attachToFile(at: requested) else {
+                    // Answering success here would leave the app believing both
+                    // processes share a file while this one keeps writing to the
+                    // previous segment — which retention eventually deletes out
+                    // from under it. Failing the request makes the app retry.
+                    diagLog.error("Listener failed to attach diagnostic logging to \(requested.path)")
+                    return nil
+                }
                 diagLog.debug("Listener attached diagnostic logging to \(requested.path)")
                 return .configureLogging
             case let .sourcePIDs(windows):
--- a/Shared/Services/MenuBarItemService.swift
+++ b/Shared/Services/MenuBarItemService.swift
@@ -15,7 +15,14 @@ nonisolated enum MenuBarItemService {
 nonisolated extension MenuBarItemService {
     enum Request: Codable {
         case start
-        case configureLogging(filePath: String)
+        /// Points the service's diagnostic logger at `filePath`, or turns its
+        /// file logging off when `nil`. Sent at startup, again on every log
+        /// rotation, and whenever the user toggles diagnostic logging.
+        ///
+        /// `rotationPolicy` carries the app's retention settings, so the
+        /// service prunes the shared log directory by the same rules rather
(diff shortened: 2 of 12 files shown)


diazdesandi:
Thanks!

Issue #759, opened 2026-07-01
Bug report: Changing Thaw settings reproducibly crashes macOS WindowServer


msnnvnr:
### Pre-submission Checklist

- [x] I searched existing issues (open & closed) and found no duplicates
- [x] I'm using the latest app version

### What happened?

Summary

Changing settings in Thaw reproducibly causes macOS WindowServer to crash. The crash does not happen simply when opening the settings window, but when interacting with or changing settings inside Thaw.

The system briefly loses the graphical session / WindowServer restarts. Two WindowServer crash reports were generated within a few minutes. Both show the same crash path through replayd, SkyLight capture APIs, IOSurface, and Metal.

One of the reports explicitly identifies Thaw as the originator:

"voucherInfos" : [{"originatorName":"Thaw","proximateName":"replayd"}]

The other report identifies replayd as the proximate process:

"voucherInfos" : [{"proximateName":"replayd"}]

Environment

* App: Thaw
* Thaw version: 2.0.0-rc.1 (47)
* Thaw update channel: Development channel enabled
* macOS: macOS 26.5.1 (25F80)
* Mac model: MacBookPro18,1
* CPU/GPU: Apple M1 Pro
* RAM: 16 GB
* Display setup at time of crash: only the built-in MacBook display was physically connected
* System Integrity Protection: enabled
* Release type: User

Additional context

The development channel was enabled in Thaw, running version 2.0.0-rc.1 (47).

BetterDisplay was installed/running around the same time, but no physical external display was attached. The crash reports show VirtualDisplayListener and external-0 / external-1 WindowServer threads, although only the internal display was connected.

This may indicate virtual or persistent display contexts, but the actual crash trigger in the reports points to replayd, and in one report specifically to Thaw via originatorName.

Steps to reproduce

1. Start Thaw.
2. Open Thaw settings.
3. Change or interact with one or more settings inside the settings UI.
4. macOS WindowServer crashes/restarts.

Expected behavior

Changing Thaw settings should not affect WindowServer or restart the graphical session.

Actual behavior

WindowServer crashes when changing/interacting with settings inside Thaw. macOS generates WindowServer crash reports.

Crash reports

Two WindowServer crash reports were created within a few minutes.

Relevant sanitized excerpts:

Crash report 1:
captureTime: 2026-07-01 09:51:38 +0200
voucherInfos: proximateName: replayd
Crash report 2:
captureTime: 2026-07-01 09:55:31 +0200
[...]

PR #1009, opened 2026-08-31
feat(appearance): Thaw Bar appearance settings and multi-display icon tint

### LINKED ISSUE #325: [Feature Request]: No border when corner shape of the Thaw bar is square

### Description

When using the square shape for the corners of the Thaw bar in combination with a border, the border is not fully visible due to the rounding of the screen:

Image

Proposed fix:
When using a border in this situation, do not apply the border of the straight edge:

Image

### Additional Information

_No response_

jiayuqi7813:
## Summary
- Give the Thaw Bar its own appearance model and editor (background, tint, corners, border, shadow), including Match Menu Bar brightness/glass sliders and #325 square-corner top-border omission.
- Paint Match Menu Bar from a live menu-bar sample, with optional brightness and liquid-glass blend, instead of sharing the menu bar overlay tint.
- Fix multi-display light/dark icon flashes by keeping per-display icon caches, restoring the warm snapshot on open, and always ordering the panel front immediately so the first click still shows the bar.
- Local Debug helpers (`scripts/setup-local-codesign.sh`, `scripts/run-debug.sh`) keep a stable self-signed identity so Accessibility / Screen Recording survive rebuilds.

PR is larger than the usual ≤500 LOC aim because the appearance model, editor, IceBar rendering, multi-display image cache, and local codesign tooling are one user-facing change set for #325 / Thaw Bar appearance.

Closes: #325

## Summary by CodeRabbit

* **New Features**
  * Added customizable Thaw Bar appearance settings, including backgrounds, tint, glass effects, corners, borders, shadows, and light/dark configurations.
  * Added live appearance previews and improved menu bar rendering across displays.
  * Added appearance-related settings search entries.

* **Bug Fixes**
  * Improved display-specific color sampling and image restoration.
  * Migrated existing appearance settings for compatibility with the new Thaw Bar options.

* **Chores**
  * Added local debug signing and launch tools.
  * Added coverage for appearance configuration, migration, encoding, and border rendering.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,8 @@ plans/
 # Local badge mockups
 .tmp-badge-preview/
 default.profraw
+
+# Local Debug codesign (self-signed cert copy + DerivedData)
+.local-codesign/
+Thaw-local/
+**/Thaw-local/
--- a/Thaw/MenuBar/Appearance/Configurations/MenuBarAppearanceConfigurationV1.swift
+++ b/Thaw/MenuBar/Appearance/Configurations/MenuBarAppearanceConfigurationV1.swift
@@ -82,6 +82,10 @@ extension MenuBarAppearanceConfigurationV2 {
             configuration.lightModeConfiguration = partialConfiguration
             configuration.darkModeConfiguration = partialConfiguration
             configuration.staticConfiguration = partialConfiguration
+            let thawBarPartial = ThawBarAppearancePartialConfiguration.migrating(from: partialConfiguration)
+            configuration.thawBarLightModeConfiguration = thawBarPartial
+            configuration.thawBarDarkModeConfiguration = thawBarPartial
+            configuration.thawBarStaticConfiguration = thawBarPartial
             configuration.shapeKind = oldConfiguration.shapeKind
             configuration.fullShapeInfo = oldConfiguration.fullShapeInfo
             configuration.splitShapeInfo = oldConfiguration.splitShapeInfo
--- a/Thaw/MenuBar/Appearance/Configurations/MenuBarAppearanceConfigurationV2.swift
+++ b/Thaw/MenuBar/Appearance/Configurations/MenuBarAppearanceConfigurationV2.swift
@@ -12,6 +12,12 @@ nonisolated struct MenuBarAppearanceConfigurationV2: Hashable {
     var lightModeConfiguration: MenuBarAppearancePartialConfiguration
     var darkModeConfiguration: MenuBarAppearancePartialConfiguration
     var staticConfiguration: MenuBarAppearancePartialConfiguration
+    /// Thaw Bar appearance for light mode when ``isDynamic`` is on.
+    var thawBarLightModeConfiguration: ThawBarAppearancePartialConfiguration
+    /// Thaw Bar appearance for dark mode when ``isDynamic`` is on.
+    var thawBarDarkModeConfiguration: ThawBarAppearancePartialConfiguration
+    /// Thaw Bar appearance when ``isDynamic`` is off.
+    var thawBarStaticConfiguration: ThawBarAppearancePartialConfiguration
     var shapeKind: MenuBarShapeKind
     var fullShapeInfo: MenuBarFullShapeInfo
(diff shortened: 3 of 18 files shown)


jiayuqi7813:
@coderabbitai review

jiayuqi7813:
@coderabbitai Thanks for the review. Addressed in the follow-up commit:

**Fixed**
- Full-height menu-bar color crop in `IceBarColorManager`
- Per-display snapshot replace + memory-pressure / size bound for `imagesByDisplay`
- Partial warm restore no longer leaves prior-display bitmaps; incomplete restore requests recapture
- Dynamic appearance editor height sums light + dark Thaw Bar bonuses
- Border shape takes circular/continuous corner style
- `recaptureSection` stamps `lastSkyLightBatchAt` for the SkyLight path
- Gradient contrast uses a location-0.5 blend
- Dropped unrestricted `-A` ACL on local codesign import

**Kept as-is**
- `ENABLE_HARDENED_RUNTIME=NO` in `scripts/run-debug.sh` — intentional for the self-signed local Debug helper (comment added)
- 1.5s Match Menu Bar refresh cadence — needed for live wallpaper tracking while the panel is open; frame/notification paths already cover discrete changes

jiayuqi7813:
image

Now he has become better-looking, rather than a very dark color (

jiayuqi7813:
@coderabbitai Follow-up for the second-round findings:

- Translucent adaptive/solid/gradient fills now composite over the sampled color (same opacity as `ThawBarChrome`) before icon contrast.
- `updateWindowImage` returns success; notification / visibility / timer / `refresh` paths skip `updateColorInfo` when the capture is stale.

diazdesandi (changes requested):
Thanks for this. I want the multi-display work and the border fix, but not as one branch. Please split it into three PRs:

**1. The #325 border fix.** `ThawBarBorderShape.swift` plus the wiring that turns off the top edge. About 74 lines. I can merge that quickly. One thing to fix first: the open-top path draws circular arcs even when `cornerStyle` is `.continuous`, so the stroke sits off the clip edge on continuous corners. The parameter is threaded through but the arc math ignores it.

**2. The multi-display fixes.** The `screen == .main` guards coming out of `IceBarColorManager`, the per-display snapshots in `MenuBarItemImageCache`, and the `orderFrontRegardless()` change. Good catches, both bugs are real. Four things before it lands:

- `recaptureSection` sets `lastSkyLightBatchAt` by hand instead of going through the rate limiter. That walks around the #759 leak throttle. Please route it through the limiter.
- The settle delay drops from 500 ms to 16 ms. The comment above it says that delay is what keeps `findSection()` from misclassifying everything as `.visible` and showing "No items…". I'd rather keep it.
- The color sample timer goes from 5s to 1.5s. What drove that number?
- `storedImage(for:in:)` reimplements the `matchesIgnoringWindowID` fallback that `image(for:)` already does. Call the existing one.

**3. The appearance model and editor.** Holding this one for now. It writes new keys into `MenuBarAppearanceConfigurationV2` and migrates existing payloads, so once a beta ships it I can't cleanly revert. I want to decide on that surface separately, not as a side effect of a border fix.

Please leave the codesign scripts out. We already have a local signing workflow, and these would need changes anyway: `security import -A` lets every process on the machine use the key, `ENABLE_HARDENED_RUNTIME=NO` overrides the Debug target, `codesign -dv` displays but doesn't verify, and `|| true` on the launch check reports success when nothing launched.

One heads up: #1046 landed and moved the Thaw Bar render path to the semantic `image(for:)` lookup. Rebase onto that before you touch `IceBar.swift`.

jiayuqi7813:
Split per review feedback:

1. **#325 border** → https://github.com/thaw-app/Thaw/pull/1064  
2. **Multi-display icon tint** → https://github.com/thaw-app/Thaw/pull/1065  
3. **Appearance model / editor** — held locally for now (writes new V2 keys + migration); not opened yet.  
Local codesign scripts are left out of both new PRs.

Closing this PR in favor of the split.

jiayuqi7813:
Superseded by #1064 and #1065; appearance work held pending separate decision.

PR #1065, opened 2026-09-07
fix(icebar): restore correct icon tint when opening on another display

jiayuqi7813:
## Summary
- Remove `screen == .main` guards in `IceBarColorManager` so secondary-display Thaw Bar opens sample the panel's own screen for icon contrast.
- Keep per-display warm icon snapshots in `MenuBarItemImageCache`, restore them before `orderFront`, and recapture through the existing offscreen SkyLight cadence (no hand-stamped bypass of the #759 throttle).
- Keep the 500 ms settle delay. Color sample timer stays at 5 s.

## Notes for review
- `storedImage` was removed in favor of a shared `image(for:in:)` helper used by both `image(for:)` and warm-restore.
- Appearance model / editor and local codesign scripts are **not** in this PR (split out per review on #1009).

## Test plan
- [ ] Dual-display: open Thaw Bar on the main screen, then on the secondary screen — icons should not flash the other display's light/dark tint
- [ ] Same-display reopen still settles without an empty "No items…" flash
- [ ] Hidden / always-hidden sections still populate after the settle delay

Closes: N/A

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved menu bar appearance when opening the panel on a different screen.
  * Prevented brief flashes of incorrect light or dark icon tints during screen changes.
  * Added a loading state while the correct screen-specific colors and icons are refreshed.
  * Preserved recently captured icon appearances per display for faster, smoother switching.

### DIFF
--- a/Thaw/MenuBar/IceBar/IceBar.swift
+++ b/Thaw/MenuBar/IceBar/IceBar.swift
@@ -34,6 +34,10 @@ final class IceBarPanel: NSPanel {
     /// change posts that notification, racing with the show).
     private var lastShowTimestamp: Date?
 
+    /// Display the Thaw Bar was last shown on. Cross-screen opens must drop
+    /// the previous screen's icon captures (light/dark tint is baked in).
+    private var lastShownDisplayID: CGDirectDisplayID?
+
     /// Storage for internal observers.
     private var cancellables = Set<AnyCancellable>()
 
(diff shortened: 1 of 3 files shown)


diazdesandi (dismissed):

diazdesandi (approved):
Approved. The two major CodeRabbit findings are fixed in 0fb66c849: `storeImages` now snapshots only the tags a capture actually produced for the display (so a one-section recapture cannot record another display's bitmaps), and `recaptureSection` returns before storing when the task was cancelled. Rebased onto current `development`; build green.

Issue #1071, opened 2026-09-09
[Feature Request] Block Asking "Apply mednu bar spacing change?" to Relaunch Apps


netheremp:
### Pre-submission Checklist

- [x] I searched existing issues (open & closed) and found no duplicates
- [x] If this is about macOS 27 / Golden Gate, I have read

### Problem / Use Case

Hello stonerl,

First of all, super appreciate for taking the baton from Ice and make it alive again.

The problem I'm encoutering right now is that Thaw constantly asking to relaunch app while this action itself can't be disabled or blocked.
My daily workflow requires plug/unplug external displays very often, while relaunching app shouldn't be processed since there are Internal VPN/Monitor App living in menu bar. So this is actually annoying when it had to ask over and over again, while I don't need it at all.
Sometimes I just open an app or recording the screen/audio then that pops up because there's MacOS internal notification on menu bar or other issues.

<img width="274" height="345" alt="Image" src="https://github.com/user-attachments/assets/45194769-b690-4111-a362-b43d2c9a3512" />

### Proposed Solution

What I only want is stop it from asking that can be disabled from setting or by checking "Don't ask again" -> Cancel, at the cost of messing up the spacing.

diazdesandi:
Fixed in #1075. Settings, Displays now has a "When applying spacing" choice. "Wait until next restart (no apps restarted)" writes the preference and leaves every app running, so the new spacing appears the next time each app starts. The "Confirm before relaunching apps" toggle and its save-scope picker only appear when a relaunch wave can actually fire.

Issue #1054, opened 2026-09-05
[Feature Request] Option to remove saved diplays in 'Per Display" section of Display Settings


lucifercraig12345-create:
### Pre-submission Checklist

- [x] I searched existing issues (open & closed) and found no duplicates
- [x] If this is about macOS 27 / Golden Gate, I have read

### Problem / Use Case

Displays connected to macvook are saved in thaw forever. After sometime, it may get cluttered. 

<img width="1470" height="508" alt="Image" src="https://github.com/user-attachments/assets/f6b79c12-406b-4173-ac16-c4b5eb86b90e" />

### Proposed Solution

Add an option to remove saved diplays in 'Per Display" section of Display Settings. As it can get cluttered after some time.

diazdesandi:
Fixed in #1075. A Remove control on each disconnected display's header in Settings, Displays drops its cached name and per-display settings, with a confirmation. A display you connect again reappears on its own.

lucifercraig12345-create:
Thank you 😄

Issue #969, opened 2026-08-24
[Bug] Search shortcut doesn't auto select typing field


quintrino:
### Pre-submission Checklist

- [x] I searched existing issues (open & closed) and found no duplicates
- [x] If this is about macOS 27 / Golden Gate, I have read
- [x] I'm using the latest app version

### What happened?

So when I hit my keyboard shortcut for search menu bar items, and start typing, it just errors out (with an alert noise to failed keyboard action), I have to manually click into the search bar window to be able to start typing. 

I assume this isn't expected behaviour? 

### Steps to Reproduce

1. Use your search menu bar items keyboard shortcut. 
2. Begin typing straight away. 

### Expected Behavior

You should be able to start typing straight away. 

/usr/local/Caskroom/thaw/1.2.0 (11.5MB)
  Installed using the formulae.brew.sh API on 2026-02-14 at 10:11:12

### App Version

1.2.0

### macOS Version

26

### Display Setup

Multiple monitors

### Logs / Console Output

[thaw_2026-08-24_10-48-37.log](https://github.com/user-attachments/files/31357121/thaw_2026-08-24_10-48-37.log)

diazdesandi:
Was this issue fixed in 2.0.1 / 2.1.0 beta?

diazdesandi:
Fixed in #1075. The search field is now first responder as soon as the panel opens, with no fixed delay to race against, so the first keystroke after the hotkey lands.

Issue #936, opened 2026-08-12
[Feature Request] Sort items in hidden section


santi-sd:
### Pre-submission Checklist

- [x] I searched existing issues (open & closed) and found no duplicates
- [x] If this is about macOS 27 / Golden Gate, I have read

### Problem / Use Case

My visible section is quite slim.

<img width="864" height="42" alt="Image" src="https://github.com/user-attachments/assets/b4612204-f429-4a6b-8848-dee86a201271" />

My hidden section is where the rest are.

<img width="987" height="78" alt="Image" src="https://github.com/user-attachments/assets/092d96c5-5792-489f-a020-831f867634d6" />

### Proposed Solution

It would be awesome if I could have the items sort themselves in alphabetical order to make it easier to find things. It would also be great if a new menu item is automatically put into my hidden section.

I was not able to find any mention of this idea in the issues search.

Thanks!

diazdesandi:
Noted, coming with 2.1.0 likely. As for the new items, there should be a New Items pill; just place it on Hidden, and that should put them there automatically.

santi-sd:
<img width="854" height="624" alt="Image" src="https://github.com/user-attachments/assets/99baac6b-8de7-4a54-bc3e-266e5eec14a9" />

I dont see a new item pill

lucifercraig12345-create:
I think it is in the newer (beta) versions. 

And honestly, you should try it out. I haven't had any major issues with it. 

(also can you give the name of the apps in your pic🙂, I know some but not all).

<img width="739" height="347" alt="Image" src="https://github.com/user-attachments/assets/7076396f-520f-4f30-ace2-b73abe79fd1c" />

santi-sd:
yes absolutely!

Asterisk* means that I've never actually used it yet.

Espanso
Paperman
Shottr
Numi
Google Drive
QSpace Pro
Tuna
KeyClu
DockDoor
Hyperkey
Syncthing
Rectangle Pro
LM Studio*
WhatCable
f.lux
Jellyfin
AltTab
Antinote

diazdesandi:
Fixed in #1075. Each section heading in Settings, Menu Bar Layout has a Sort A→Z button that reorders the section by display name, keeps identifiers for closed apps via planSectionOrder's merge, and writes the order to the active profile.

PR #1075, opened 2026-09-09
fix(menubar): stop layout and spacing work from fighting the user

### LINKED ISSUE #1055: [Bug] Thaw becomes unresponsive after typing in the search bar and scrolling
### What happened?

Whenever I type something in the thaw app search bar and scroll, the app becomes unresponsive and I have to force quit it. 

### Steps to reproduce

In thaw app type anything in the search bar and scroll. 

### Logs, screenshots, or a screen recording

[thaw_2026-09-06_02-20-47.log](https://github.com/user-attachments/files/31870543/thaw_2026-09-06_02-20-47.log)
[thaw_2026-09-06_02-21-51.log](https://github.com/user-attachments/files/31870545/thaw_2026-09-06_02-21-51.log)
[thaw_2026-09-06_02-18-52.log](https://github.com/user-attachments/files/31870544/thaw_2026-09-06_02-18-52.log)
[thaw_2026-09-06_02-16-29.log](https://github.com/user-attachments/files/31870546/thaw_2026-09-06_02-16-29.log)

### Thaw version

Version 2.1.0-beta.2 (58)

### macOS version

26.6.2

### Display setup

None

### Anything else?

also getting this a lot when moving items. like almost always. 

Image

### LINKED ISSUE #1056: [Bug] App crashed after adding spacer to visible section
### What happened?

App crashed after adding spacer to visible section. App wont even open. 

Whenever I open Thaw getting this - 

Image

Also cannot provide log as unable to open the app. 

### Steps to reproduce

Add spacer to visible section

### Logs, screenshots, or a screen recording

[SCR-20260906-ddxw.png](https://github.com/user-attachments/assets/825fe559-9dfa-4a54-8816-936a051de6d0)

### Thaw version

Version 2.1.0-beta.2 (58)

### macOS version

26.6.2

### Display setup

None

### Anything else?

I don't know if I should attach the apple report here, as I am unsure of how private it is? Please tell me if it's alright to attach it.

### LINKED ISSUE #1058: [Bug] dropReverted error encountered on Thaw 2.1.0b2
### What happened?

Upon updating AltTab to version 11.6.0, Thaw 2.1.0b2 encountered an exception.

### Steps to reproduce

1. Terminate the AltTab application.
2. Overwrite the existing AltTab 11.6.0 executable via Finder.
3. Initialise AltTab version 11.6.0.
4. Observe the resultant error dialogue presented by Tahw 2.1.0b2.

### Logs, screenshots, or a screen recording

[Thaw-move-diagnostic-2026-09-06-080210.txt](https://github.com/user-attachments/files/31872049/Thaw-move-diagnostic-2026-09-06-080210.txt)

### Thaw version

2.1.0b2

### macOS version

macOS Tahoe.6.2

### Display setup

Single display

### Anything else?

_No response_

### LINKED ISSUE #1060: Status items never rendered: hiddenWindowNumber has zero low 32 bits, no CGWindowID ever allocated (2.1.0-beta.2)
## Summary

Thaw places no status items at all. The menu bar is byte-for-byte identical whether Thaw is running or quit. The process stays alive and healthy, but `ControlItem.Hidden` is reported as occluded at every launch and the control-item lookup then fails forever.

This has persisted since 2026-08-22 across **1.2.0**, **2.0.0-rc.5** and now **2.1.0-beta.2 (58)**.

## Environment

- Thaw 2.1.0-beta.2 (58), clean install from the official DMG
- macOS 26.5.1 (25F80)
- MacBook Air (M1, `MacBookAir10,1`), 8 GB
- Built-in Retina display only, 2560x1600 (1440 pt wide). Also reproduced with an external display attached.
- Diagnostic logging enabled

## What the log shows

At startup:

```
[WARNING] [ControlItem] Thaw.ControlItem.Hidden is occluded — the menu bar accepted the status item but is not rendering it
```

then, every ~600 ms, forever:

```
[DEBUG] [MenuBarItemManager] ControlItemPair: strategy 4 (AX frame) found no confident correlation among 15 candidate item(s)
[WARNING] [MenuBarItemManager] ControlItemPair: unresolved; no strategy identified the hidden control item among 15 item(s)
[WARNING] [MenuBarItemManager] cacheItemsRegardless: Missing control item for hidden section (expected tag: com.stonerl.Thaw:Thaw.ControlItem.Hidden), keeping last-known-good cache. Items remaining: 15, windowIDs: 15. hiddenWindowNumber=17179869184, hiddenControlItemWID=nil, alwaysHiddenWindowNumber=21474836480, alwaysHiddenControlItemWID=nil. consecutiveFailures=117
```

## The part that looks diagnostic

The reported window numbers are exact multiples of 2^32:

- `hiddenWindowNumber=17179869184` = `0x4_00000000` (4 x 2^32)
- `alwaysHiddenWindowNumber=21474836480` = `0x5_00000000` (5 x 2^32)

The low 32 bits — the actual `CGWindowID` — are **zero** in both cases. So it does not look like Thaw failing to *match* its control item among the other windows: macOS appears never to have allocated a window for the status item in the first place. `getMenuBarItemWindows` consistently returns 15 windows and Thaw's own two are not among them, which is consistent with that reading.

If that is right, no pairing strategy can succeed, and `consecutiveFailures` will climb indefinitely (it reached 1431 on the previous build before I restarted).

## Feedback on 2.1.0-beta.2

The orphaned-Control-Center fix in beta.2 **did** work here, and it is a real improvement — worth knowing since it was not what I was chasing. On 2.0.0-rc.5 the log was full of:

```
getMenuBarItems: created 15 items, 15 with nil sourcePID: <com.apple.controlcenter:Item-0:4 (windowID: 32211)>, <com.apple.controlcenter:Amphetamine (windowID: 28564)>, ...
```

Third-party items were all attributed to `com.apple.controlcenter`. On beta.2 that is gone:

```
getMenuBarItems: created 15 items, all with resolved sourcePID
```

The occlusion problem is unaffected by that fix, so I believe they are two distinct bugs.

## Already ruled out

- **Permissions** — log reports `permissions state = hasAll, accessibility = true, screenRecording = true`; both green in Thaw's own Permissions window.
- **Corrupted preferences** — preference domain emptied completely, identical behaviour.
- **Menu bar saturation** — tested down to 8 items in the bar (quitting Stats, Bitwarden, CopyClip, Dropover, BetterDisplay and others); still occluded.
- **External display** — reproduces on the built-in display alone. `spans-displays = 0`.
- **Full restart** — rebooted, and the log written 35 minutes later shows the same occlusion.
- `killall SystemUIServer` — no effect.
- **Stale bundle** — only one Thaw.app on disk; `lsregister -f` made no difference. Signature valid and notarised.
- No crashes at any point.

## Anything else useful?

Happy to send the complete diagnostic log, or to run any instrumented build or extra logging that would help pin down whether the status item window is ever created. Thanks for the very detailed logging in the 2.x builds — it is what made this narrowable at all.

### LINKED ISSUE #1070: [Bug] MenuBarItemSpacingManager causes mass application restarts and crashes system services (Spotlight, TextInputMenuAgent) on display wake / sleep
### Describe the Bug
When waking from sleep, connecting/disconnecting an external display, or changing laptop clamshell state, Thaw's `MenuBarItemSpacingManager` / `applyOffset` mechanism triggers an aggressive verification loop across all menu bar items.

During this process:
1. **Third-party applications (e.g., Notion, utilities) are abruptly terminated and restarted** because Thaw flags their status items as `missing post-wave` and executes an unwanted relaunch cascade.
2. **System services (`com.apple.Spotlight`, `com.apple.TextInputMenuAgent`) crash permanently** because macOS Launch Constraints (AMFI) strictly block external apps from spawning CoreServices binaries (`SIGKILL (Code Signature Invalid) / Launch Constraint Violation`). As a result, Spotlight dies completely and `Cmd + Space` stops working until restarted manually via `launchctl`.

### Environment
- **App Version:** Thaw 2.0.1-rc.2 (Build 55) *(also reproduced on 2.0.0-rc.5 and 2.0.1)*
- **macOS Version:** macOS 26.6.2 (Build 25G83)
- **Architecture:** Apple Silicon (Apple M3)

### Steps to Reproduce
1. Run Thaw 2.0.1 with several applications running in the menu bar (e.g., Notion, Spotlight, third-party utilities).
2. Trigger a display or power state change (close/open MacBook lid in clamshell mode, plug in an external monitor, or wake from sleep).
3. Observe all running menu bar apps restarting simultaneously and losing state.
4. Try using Spotlight (`Cmd + Space`) — it fails to open due to process termination.

### Expected Behavior
Thaw should only adjust icon positions/visibility visually. It should never send quit or restart signals to running applications or interact with system CoreServices.

### Actual Behavior
- All running menu bar apps are killed and restarted in a domino effect.
- System services fail to relaunch and remain dead with Launch Constraint violations.

### Unified Logs & Crash Evidence
**1. Thaw killing and attempting to spawn applications:**
```log
[com.stonerl.Thaw:MenuBarItemSpacingManager] applyOffset verification: 1 app(s) missing post-wave: com.apple.Spotlight — running fallback
[com.stonerl.Thaw:MenuBarItemSpacingManager] applyOffset fallback NSWorkspace.openApplication(at: /System/Library/CoreServices/Spotlight.app) for com.apple.Spotlight failed: Error Domain=NSCocoaErrorDomain Code=256 "The application “Spotlight.app” could not be launched because a miscellaneous error occurred." UserInfo={NSURL=file:///System/Library/CoreServices/Spotlight.app/, NSLocalizedDescription=The application “Spotlight.app” could not be launched because a miscellaneous error occurred., NSUnderlyingError=0xbac1e0810 {Error Domain=RBSRequestErrorDomain Code=5 "Launch failed." UserInfo={NSLocalizedFailureReason=Launch failed., NSUnderlyingError=0xbac1e11d0 {Error Domain=NSPOSIXErrorDomain Code=162 "Unknown error: 162" UserInfo={NSLocalizedDescription=Launchd job spawn failed}}}}}
[com.stonerl.Thaw:MenuBarItemSpacingManager] applyOffset fallback verification: com.apple.Spotlight still missing
[com.stonerl.Thaw:MenuBarItemSpacingManager] applyOffset: 1 app(s) failed to relaunch: com.apple.Spotlight
```

**2. AMFI terminating CoreServices on spawn attempt:**
```log
kernel: (AppleMobileFileIntegrity) AMFI: Launch Constraint Violation (enforcing), error info: c[1]p[1]m[1]e[1], (Constraint not matched) launching proc[vc: 1 pid: 68209]: /System/Library/CoreServices/Spotlight.app/Contents/MacOS/Spotlight, launch type 3
```

**3. Corresponding crash reports generated in `~/Library/Logs/DiagnosticReports/`:**
- `Notion-*.ips` (abruptly terminated)
- `Spotlight-*.ips` (terminated with `CODESIGNING: Launch Constraint Violation`)
- `TextInputMenuAgent-*.ips` (terminated with `CODESIGNING: Launch Constraint Violation`)

### LINKED ISSUE #1072: [Bug] Searching for system menu bar items and pressing Enter opens the right-click menu.
### What happened?

Searching for system menu bar items (Bluetooth, Sound, WiFi, Music Recognition) and pressing Enter opens the right-click menu instead of the item. Searching for third-party apps (Raycast, Cleanshot, Bettermouse) works as expected. Affects Thaw 2.0.0 (build 53) and 2.0.1 on macOS 26.6.2.

### Steps to reproduce

N/A

### Logs, screenshots, or a screen recording

[CleanShot_2026-09-09_at_04.41.17.webp](https://github.com/user-attachments/assets/16d495a5-2ae3-45f9-a870-1692e142f2e2)

### Thaw version

2.0.0 / 2.0.1

### macOS version

26.6.2

### Display setup

None

### Anything else?

_No response_

diazdesandi:
# Summary

Twelve commits against the menu bar layout, move, and spacing paths. Five reported bugs, four unreported ones found while fixing those, and three performance changes to the image capture path that the layout work kept tripping over.

The one worth reading about is #1070. Applying menu bar spacing has to restart the apps that own status items, because macOS reads `NSStatusItemSpacing` when an owner process starts and never again. That restart wave was indiscriminate. It terminated launch-constrained system binaries too, AMFI killed the respawn at exec, and Spotlight stayed dead until reboot with `Cmd+Space` gone with it. The wave stays, since instant spacing depends on it. It now triages first.

**Scope:** larger than the usual single fix. These commits share the same three files in the move and layout paths, and splitting them means either stacked PRs on top of each other or conflicting ones. 

## Linked issue (required)

Closes: #1070
Closes: #1055
Closes: #1056
Closes: #1058
Closes: #1060

## PR Type

- [x] Bug fix
- [ ] CI/CD
- [ ] Documentation
- [ ] Feature
- [ ] Enhancement
- [x] Performance improvement
- [ ] Refactor
- [x] Test addition or update
- [ ] Other (please describe)

## Area

- [x] menubar
- [ ] icebar
- [x] layout
- [ ] appearance
- [x] settings
- [ ] onboarding
- [ ] permissions
- [ ] profiles
- [ ] hotkeys
- [ ] updates
- [ ] ops

- [ ] Yes
- [x] No

One behavior change worth calling out: after a spacing change, a few menu bar items keep their old spacing until their app next starts on its own. That is the trade for not killing Spotlight. `FREQUENT_ISSUES.md` gains a section explaining which apps Thaw will and won't quit, so the asymmetry is documented rather than surprising.

### Spacing no longer quits what it cannot bring back (#1070)

`SpacingRelaunchPolicy` classifies every candidate before anything is signalled:

1. Indexed LaunchAgent, so there's a label to restart with: `launchctl kickstart -k gui/<uid>/<label>`. launchd stays the launching parent, which is the only route a constrained binary accepts.
2. Under a system path prefix with no label: left running. Terminating it is one way, since `terminate()` is a successful exit and a `KeepAlive.SuccessfulExit=false` agent is never respawned.
3. No launchable bundle, such as XPC helpers and extension hosts: left running.
4. Everything else: quit and launched back, same as before.

`forceTerminate` is gone with it. An app that declines the quit request keeps running now, so an open save sheet no longer costs the user their work for the sake of an evenly spaced menu bar. Anything left running is dropped from the expected-reattachment set, so the settling period stops waiting on items that never detached.

The reporter's specific Spotlight crash was already fixed on `development` by #1029. What remained were the unclaimed system binaries that no agent index covers, and the `forceTerminate` on third-party apps.

### Layout work stops fighting the pointer

Three separate ways an automatic batch could stomp on what the user was doing:

1. An explicit profile apply, a background re-sort, and a saved-order restore could run at once and write over each other. Each claims a lease ranked by authority now. A profile the user selected supersedes background work and never the reverse, and a superseded batch stops at its next move instead of finishing on top of the new one.
2. Automatic batches moved items out from under a pointer that was still in use. A batch waits for a pause in physical input before starting, checks again between moves, and defers the rest when input resumes.
3. Cursor restoration warped the pointer back to wherever a batch started, even when the user had moved the mouse mid-batch. It only happens now when HID timestamps show no physical input since the operation took ownership.

### Moves resolve against live windows

`refreshMoveEndpoints` re-reads the exact windows a move uses instead of leaning on persisted PID seeds and title-matched ownership, so a failed live resolution rejects the move rather than dressing up a stale endpoint as current. `moveEndpointDisposition` checks parked lane geometry before trusting the `isOnScreen` bit, which is what turned post-update moves into `dropReverted` (#1058). `CGWindowID(exactly:)` accepted any value that fit, so synthetic and negative window numbers became plausible-looking IDs; `windowServerID(windowNumber:)` rejects them (#1060).

`AXShowMenu` is no longer tried first on Apple's own status items. They read it as a contextual-menu request rather than a click, so activation opened the wrong thing. Only `AXPress` is used.

### Crash and freeze fixes

`StatusItemStorage` asked AppKit for a zero-length status item, which can stay a synthetic window whose `windowNumber` has no representable `CGWindowID`. Resizing that crashed. It starts one point wide and the first `updateStatusItem` pass applies the real control-item length (#1056).

Every row in `SectionedList` carried a `GeometryReader` reporting its frame upward, so each keystroke re-measured the whole list and the settings search field froze. Frame tracking is gone and scroll-to-selection uses a center anchor (#1055).

### Performance

`refreshImages` was only `nonisolated`, which under Approachable Concurrency keeps the caller's actor, so bounds queries, cropping, and detached copies all ran on the main thread next to UI work. `@concurrent` moves them to the background pool and publication hops back through `applyRefreshedImages`.

Attention detection was a single Boolean demand, so one enabled blink trigger made the live loop capture every item in the concealed sections. The image cache takes the identifiers the enabled triggers name and captures only those windows, in one batched request instead of one round trip each. The global "surface items seeking attention" setting still samples everything.

## PR Checklist

- [x] I've built and run the app locally and verified that it works as expected.
- [x] I've updated documentation as needed.
- [x] If this PR changes dependencies / lockfiles, `dependency-sca` is green. (No dependency changes.)

Test commands run:

- `xcodebuild test -project Thaw.xcodeproj -scheme Thaw -destination 'platform=macOS' CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO`

3281 tests pass. New coverage: `SpacingRelaunchPolicyTests` (10 tests over the triage, including the AMFI case and path-prefix lookalikes such as `/Systems/` and `/usrlocal/`), plus `MoveEndpointIdentityTests`, `MoveEventCoordinatesTests`, `BulkMoveCoordinationTests`, and `MenuBarItemImageCacheGatingTests`.

- Items whose owners are left running keep the previous spacing until those apps next start. Documented in `FREQUENT_ISSUES.md` rather than worked around.
- The system path prefix list (`/System/`, `/usr/`, `/bin/`, `/sbin/`, `/Library/Apple/`) is a heuristic. An app installed outside those paths that still carries launch constraints would be misclassified as relaunchable.

## Other information

28 files, +1364/-348. Roughly two thirds of the insertions are tests and the new `SpacingRelaunchPolicy` file.

## Summary by CodeRabbit

- **New Features**
  - Added “Sort A→Z” for menu bar sections.
  - Added a setting to apply spacing changes immediately or at the next restart.
  - Added confirmation and cleanup options for removing saved displays.
  - Search now focuses more reliably when opened.

- **Improvements**
  - Spacing changes avoid restarting protected macOS services and handle eligible apps more safely.
  - Stale relaunch states expire after one day.
  - Menu bar activation, multi-monitor recovery, attention indicators, and image refreshes are more reliable.
  - Automatic layout changes defer when physical input is detected.

- **Documentation**
  - Expanded guidance for multi-monitor transitions, spacing behavior, and restarting affected system apps.


leos:
My two cents here are that even picking which apps can restart isn't a great approach. Other apps that have allowed folks to change spacing (Ice, and others that I've tried) have required folks to restart the whole system to have the menu spacing take effect. While that's annoying when you're trying to set it up, my guess is that most people set it once and forget about it and don't actually take advantage of being able to have different spacing on different monitors.

Any sort of app (even ones that appear 'safe') restarts on external monitors being connected/disconnected are really unexpected and could cause problems.

diazdesandi:
We added warnings everywhere. The relaunch is to allow users to have that without restarting, and it seems that both approaches are an issue. I guess I could add a toggle to let the users select the behavior they prefer.

leos:
> We added warnings everywhere. The relaunch is to allow users to have that without restarting, and it seems that both approaches are an issue. I guess I could add a toggle to let the users select the behavior they prefer.

If there was a way to disable the app restarts I definitely would. For a while I didn't understand why everything kept crashing and restarting when I was plugging into a monitor, and especially when I had to plug in and unplug several times in a row, then my whole mac was a mess and disrupted me setting up my presentation pretty significantly.

diazdesandi:
This PR also closes, via the commits added after the original body:

- Closes #1054 — remove saved displays from the Per Display list
- Closes #1071 — option to apply spacing without restarting apps
- Closes #1079 — age out stuck waitForRelaunch sentinels
- Closes #1080 — prune localized Control Center ghost identifiers
- Closes #936 — sort a section alphabetically
- Closes #969 — search field takes the first keystroke

